PR #3, opened 2018-07-12
use approximate equality with 1 ULP of tolerance

japaric:


japaric:
Workaround for the failure in https://travis-ci.org/japaric/libm/jobs/403004781#L573

PR #61, opened 2018-07-13
Implement `round`

### LINKED ISSUE #31: implement round
Source: https://git.musl-libc.org/cgit/musl/tree/src/math/round.c

notriddle:
Fixes #31

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -12,12 +12,6 @@
 #![deny(warnings)]
 #![no_std]
 
-macro_rules! force_eval {
-    ($e:expr) => {
-        unsafe { ::core::ptr::read_volatile(&$e); }
-    }
-}
-
 mod math;
 
 #[cfg(todo)]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -12,6 +12,12 @@
 #![deny(warnings)]
 #![no_std]
 
+macro_rules! force_eval {
+    ($e:expr) => {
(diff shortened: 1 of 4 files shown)


japaric:
Thanks for the PR, @notriddle.

I'm changing the structure of this crate in PR #62 to make integration of this crate into compiler-builtins easier. I'll have to ask you to rebase this PR once #62 lands. Sorry for the trouble.

src/lib.rs
```diff
@@ -26,6 +28,9 @@ pub fn _eq(a: u64, b: u64) -> bool {
     (a as i64).wrapping_sub(b as i64).abs() <= 1
 }
 
+#[inline(never)]
+fn force_eval<T>(_: T) {}
```
japaric: I believe the compiler is still free to opt away a function call to this function because it has no side effects and doesn't modify the input. I think something closer to the MUSL macro would be something like this:

``` rust
#[inline(always)]
fn force_eval<T>(x: &T) {
    unsafe { ptr::read_volatile(x); }
}
```

And you would invoke it like this: `force_eval(&(x + TOINT));`

japaric:
Looks great! Thanks @notriddle.

bors r+

notriddle:
Erm, no. For some reason, my macro doesn't work with the `cb` thing. Sorry.

notriddle:
There, now you'll need to re-r+ it.

PR #66, opened 2018-07-13
add cosf with dependencies

burrbull:


src/math/service/rem_pio2_large.rs
```diff
@@ -0,0 +1,489 @@
+use ::scalbn;
```
japaric: this needs to be a relative import

src/math/service/rem_pio2_large.rs
```diff
@@ -0,0 +1,489 @@
...
+use ::F64Ext;
```
japaric: you can't use anything from outside the `math` module. You'll have to use free functions instead of methods.

src/math/mod.rs
```diff
@@ -15,10 +15,11 @@ mod sqrtf;
 mod logf;
 mod expf;
 mod floor;
+mod cosf;
 mod trunc;
 mod truncf;
 
-//mod service;
+mod service;
```
japaric: (we don't *need* to put helpers in a module because we have the concept of privacy in Rust so these helpers could be files in src/math.rs -- the important part is not make them public when we re-export them in the math module (i.e. use `use ..` or maybe `pub(crate) use ..` instead of `pub use ..`)

PR #67, opened 2018-07-13
cosf with corrections from review

burrbull:


src/math/rem_pio2_large.rs
```diff
@@ -0,0 +1,489 @@
+use super::scalbn;
+use super::floor;
+
+/// double x[],y[]; int e0,nx,prec;
```
japaric: these can be turned into plain comments (`//` instead of `///`). Doc comments won't be rendered for this function / const / item because it's private.

src/math/rem_pio2_large.rs
```diff
@@ -0,0 +1,489 @@
...
+    0x3991D6, 0x398353, 0x39F49C, 0x845F8B, 0xBDF928, 0x3B1FF8,
+    0x97FFDE, 0x05980F, 0xEF2F11, 0x8B5A0A, 0x6D1F6D, 0x367ECF,
+    0x27CB09, 0xB74F46, 0x3F669E, 0x5FEA2D, 0x7527BA, 0xC7EBE5,
+    0xF17B3D, 0x0739F7, 0x8A5292, 0xEA6BFB, 0x5FB11F, 0x8D5D08,
+    0x560330, 0x46FC7B, 0x6BABF0, 0xCFBC20, 0x9AF436, 0x1DA9E3,
+    0x91615E, 0xE61B08, 0x659985, 0x5F14A0, 0x68408D, 0xFFD880,
+    0x4D7327, 0x310606, 0x1556CA, 0x73A8C9, 0x60E27B, 0xC08C6B,
+];
+
+#[cfg(ldbl_max_exp_more1024)]
```
japaric: This table will always be ignored by the compiler because `--cfg ldbl_max_exp_more1024` is never passed to rustc. I think we want to change the contents of this table depending on the architecture. At least that's what the MUSL code does:

``` console
$ rg 'define.*LDBL_MAX_EXP'
arch/x32/bits/float.h
14:#define LDBL_MAX_EXP 16384

arch/sh/bits/float.h
10:#define LDBL_MAX_EXP 1024

arch/powerpc64/bits/float.h
10:#define LDBL_MAX_EXP 1024

arch/powerpc/bits/float.h
10:#define LDBL_MAX_EXP 1024

arch/or1k/bits/float.h
10:#define LDBL_MAX_EXP 1024

arch/mipsn32/bits/float.h
10:#define LDBL_MAX_EXP 16384

arch/microblaze/bits/float.h
10:#define LDBL_MAX_EXP 1024

arch/mips64/bits/float.h
10:#define LDBL_MAX_EXP 16384

arch/i386/bits/float.h
14:#define LDBL_MAX_EXP 16384

arch/mips/bits/float.h
10:#define LDBL_MAX_EXP 1024

arch/arm/bits/float.h
10:#define LDBL_MAX_EXP 1024

arch/aarch64/bits/float.h
10:#define LDBL_MAX_EXP 16384

arch/s390x/bits/float.h
10:#define LDBL_MAX_EXP 16384

arch/x86_64/bits/float.h
14:#define LDBL_MAX_EXP 16384
```

It seems that the smaller table is for 32-bit architectures and the bigger one is for 64-bit architectures. So you probably want to use `#[cfg(target_pointer_width = "32")]` on the smaller table and  `#[cfg(target_pointer_width = "64")]` on the bigger one.

japaric:
The cosf test is failing with different inputs: https://travis-ci.org/japaric/libm/builds/403804643

src/math/rem_pio2_large.rs
porglezomp: I believe this is actually based on `long double` sizes, where it's larger for architectures that support 80-bit or 128-bit floats. I don't think Rust supports larger than `f64`, so I don't think we need these.

porglezomp:
I commented on an outdated comment, so I'm not sure if it shows up for you, but replying to the comment about the `cfg`:

> It seems that the smaller table is for 32-bit architectures and the bigger one is for 64-bit architectures.

I believe this is actually based on `long double` sizes, where it's larger for architectures that support 80-bit or 128-bit floats. I don't think Rust supports larger than `f64`, so I don't think we need these?

porglezomp:
As discussed on #76, I propose renaming `cosdf` to `k_cosf` since that appears to be the original name in the FreeBSD source, and for symmetry with my `__cos` becoming `k_cos`.

burrbull:
Ok.

src/math/rem_pio2_large.rs
```diff
@@ -0,0 +1,483 @@
...
+pub(crate) fn rem_pio2_large(x : &[f64], y : &mut [f64], e0 : i32, prec : usize) -> i32 {
+    let x1p24  = f64::from_bits(0x4170000000000000); // 0x1p24 === 2 ^ 24
+    let x1p_24 = f64::from_bits(0x3e70000000000000); // 0x1p_24 === 2 ^ (-24)
+    
+    #[cfg(target_pointer_width = "64")]
+    assert!(e0 <= 16360);
+    
+    let nx = x.len();
+    
+    let mut fw : f64;
```
japaric: this file needs to be `cargo fmt`-ed.

japaric:
@porglezomp in that case we should shrink the table once we have more implementations that depend on rem_pio_large in tree. Then we can verify (via the test suite) that shrinking the table doesn't break existing tests.

bors r+

PR #71, opened 2018-07-13
implement sqrt and hypot

### LINKED ISSUE #22: implement hypot
Source: https://git.musl-libc.org/cgit/musl/tree/src/math/hypot.c

### LINKED ISSUE #57: implement sqrt
Source: https://git.musl-libc.org/cgit/musl/tree/src/math/sqrt.c

erikdesjardins:
closes #57, closes #22

### DIFF
--- a/src/math/hypot.rs
+++ b/src/math/hypot.rs
@@ -1,0 +1,74 @@
+use core::f64;
+
+use super::sqrt;
+
+const SPLIT: f64 = 134217728. + 1.; // 0x1p27 + 1 === (2 ^ 27) + 1
+
+fn sq(x: f64) -> (f64, f64) {
+    let xh: f64;
+    let xl: f64;
+    let xc: f64;
+
+    xc = x * SPLIT;
+    xh = x - xc + xc;
+    xl = x - xh;
+    let hi = x*x;
+    let lo = xh*xh - hi + 2.*xh*xl + xl*xl;
+    (hi, lo)
+}
+
+#[inline]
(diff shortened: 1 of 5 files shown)


src/math/hypot.rs
```diff
@@ -0,0 +1,74 @@
+use core::f64;
+
+use super::sqrt;
+
+const SPLIT: f64 = 134217728. + 1.; // 0x1p27 + 1 === (2 ^ 27) + 1
```
erikdesjardins: I'm not sure if this is correct; the source is:
```c
#if FLT_EVAL_METHOD > 1U && LDBL_MANT_DIG == 64
#define SPLIT (0x1p32 + 1)
#else
#define SPLIT (0x1p27 + 1)
#endif
```
I don't fully understand `FLT_EVAL_METHOD`, but I believe that Rust semantics are equivalent to `#define FLT_EVAL_METHOD 0`

japaric: Both `FLT_EVAL_METHOD` and `FLT_EVAL_METHOD` are defined in the MUSL source code (`grep -r 'define.*FLT_EVAL_METHOD`) and their value depends on the target architecture. Though it seems that for all the architectures that Rust supports the if condition will eval to false so this is good as it is.

If that were not the case this would need cfg attributes of this sort: `#[cfg(any(target_arch = "x86_64", target_arch = "aarch64")]`.

japaric:
Thanks @erikdesjardins 

bors r+

Issue #56, opened 2018-07-12
implement floorf


japaric:
Source: https://git.musl-libc.org/cgit/musl/tree/src/math/floorf.c

jackmott:
on it

japaric:
Done in #77

Issue #54, opened 2018-07-12
implement ceilf


japaric:
Source: https://git.musl-libc.org/cgit/musl/tree/src/math/ceilf.c

jackmott:
on it.

japaric:
Done in #77

PR #77, opened 2018-07-14
adding ceilf and floorf 

jackmott:
Finishes issues #56  and #54


japaric:
Thanks @jackmott!

bors r+

PR #84, opened 2018-07-14
Implement fmod

### LINKED ISSUE #21: implement fmod
Source: https://git.musl-libc.org/cgit/musl/tree/src/math/fmod.c

P1n3appl3:
closes #21 
I replaced the `isnanf()` function from `fmodf()` with a call to the core function `is_nan()`. If there's a reason we needed to re-implement it then I can change that back.

Also I couldn't figure out what to do in `src/lib.rs`. Is the `#[cfg(todo]` that I should be removing the one for `mod_euc()`? If so why is the equivalent one for the `f64` version still there?


japaric:
> I replaced the isnan() function from fmodf() with a call to the core function is_nan(). If there's a reason we needed to re-implement it then I can change that back.

That's OK. I didn't notice that f32 / f64 had *some* methods available in core.

> Also I couldn't figure out what to do in src/lib.rs. Is the #[cfg(todo] that I should be removing the one for mod_euc()? If so why is the equivalent one for the f64 version still there?

Nothing needs to be done in src/lib.rs for fmodf. None of the methods directly uses fmodf or fmod.

I believe the compiler will inject calls to `fmodf` when lowering expressions like `x % y` where both x and y have type f32. If that's the case then we'll probably need to add #[no_mangle] symbols for fmodf and fmod to this library. I'll check if that's the case and do a follow up PR if necessary.

This LGTM. Thanks @P1n3appl3!

bors r+

P1n3appl3:
not sure how that happened, i didn't even touch `log2` but it failed the tests

src/math/mod.rs
```diff
@@ -31,12 +32,8 @@ mod truncf;
 //mod service;
 
 pub use self::{
-    ceilf::ceilf, expf::expf, fabs::fabs, fabsf::fabsf, floor::floor, floorf::floorf, fmodf::fmodf,
-    hypot::hypot, hypotf::hypotf, log2::log2, log2f::log2f, logf::logf, powf::powf, round::round,
```
japaric: the log2 and log2f imports got removed here. Maybe due to some rebase? Please add them back

src/lib.rs
```diff
@@ -229,7 +229,7 @@ impl F32Ext for f32 {
 
     #[inline]
     fn log2(self) -> Self {
-        log2f(self)
+        self.log2f(self)
```
japaric: Undo this. This will not compile.

src/lib.rs
```diff
@@ -556,7 +556,7 @@ impl F64Ext for f64 {
 
     #[inline]
     fn log2(self) -> Self {
-        log2(self)
```
japaric: Undo this. This will not compile.

src/math/mod.rs
P1n3appl3: ah that was it thanks

P1n3appl3:
Maybe in travis running rust-fmt before the actual tests would be good so it doesn't do all that work if it's just going to fail.

japaric:
You can argument for the opposite as well: if it does the formatting first and fails then the implementation never gets tested. I think a compromise would be to make the *first* build job run *both* the tests and try to format the code and have it fail if any of the two failed; then we can enable the fast_finish feature in .travis.yml.

This needs to be rebased. After rebasing please leave a comment that contains the text "bors r=japaric" without the quotes to put this PR in the bors queue.

bors delegate+

japaric:
bors retry

Failed due to network errors

PR #91, opened 2018-07-14
Implement exp2 and exp2f

### LINKED ISSUE #15: implement exp2
Source: https://git.musl-libc.org/cgit/musl/tree/src/math/exp2.c

### LINKED ISSUE #16: implement exp2f
Source: https://git.musl-libc.org/cgit/musl/tree/src/math/exp2f.c

porglezomp:
Closes #15
Closes #16


src/math/exp2.rs
```diff
@@ -0,0 +1,383 @@
...
+    let p1 = f64::from_bits(0x3fe62e42fefa39ef);
+    let p2 = f64::from_bits(0x3fcebfbdff82c575);
+    let p3 = f64::from_bits(0x3fac6b08d704a0a6);
+    let p4 = f64::from_bits(0x3f83b2ab88f70400);
+    let p5 = f64::from_bits(0x3f55d88003875c74);
+
+    #[cfg_attr(rustfmt, rustfmt_skip)]
+    let tbl = [
+        //  exp2(z + eps)                   eps
+        f64::from_bits(0x3fe6a09e667f3d5d), f64::from_bits(0x3d39880000000000),
```
japaric: Let's DRY this and store the table as `[u64]`. The conversion to `f64` can be done after indexing the table.

elichai:
What does rust promise when it comes to f32 < MIN_EXP with exp2 or f32 > MAX_EXP with exp2?

because if it doesn't promise anything can't we just do this?
`f32::from_bits(((127i32 + exp) as u32) << 23u32)`
(subtracting the exponent from E Max and shifting to precision -1)

Lokathor:
Our general promise for the crate is to follow the C/C++ math spec as closely as we can.

In this case: https ://en.cppreference.com/w/c/numeric/math/exp2

I don't have particular knowledge of this function, but it's probably the same as normal.

PR #92, opened 2018-07-14
Use separate imports instead of {}-grouped imports for better merges

porglezomp:
Previously every merge would cause conflicts because the {}-group would re-flow, which the merge algorithm can't handle. This will hopefully make rebases and merges go more smoothly while everyone is still adding new modules.


japaric:
Please "bors r=japaric" this after PR #67 lands as it will need to be rebased.

bors delegate+

PR #95, opened 2018-07-14
add an implemetation of asin

### LINKED ISSUE #7: implement asin
Source: https://git.musl-libc.org/cgit/musl/tree/src/math/asin.c

vjackson725:
Fixes #7

Additionally adds inline functions for macros dealing with low and high words of f64s to the root module.
I'm not sure if you want these, but they seem sufficiently different to `get_bits` to merit a separate function.

Tell me if I need to change anything!

### DIFF
--- a/src/math/mod.rs
+++ b/src/math/mod.rs
@@ -6,6 +6,7 @@ macro_rules! force_eval {
     };
 }
 
+mod asin;
 mod ceilf;
 mod expf;
 mod fabs;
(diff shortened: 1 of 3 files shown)


vjackson725:
Whoops, I misinterpreted the hex floating point literal as a subtraction. Fixed now.

src/math/mod.rs
```diff
@@ -36,12 +37,39 @@ mod truncf;
...
+    (x.to_bits() >> 32) as u32
+}
+
+#[inline]
+pub fn get_low_word(x: f64) -> u32 {
+    x.to_bits() as u32
+}
+
+#[inline]
+pub fn with_set_high_word(f: f64, hi: u32) -> f64 {
```
japaric: I think it would be more rustic to write this as `fn set_high_word(f: &mut f64, hi: u32)`.

src/math/mod.rs
```diff
@@ -36,12 +37,39 @@ mod truncf;
...
+    let mut tmp = f.to_bits();
+    tmp &= 0x00000000_ffffffff;
+    tmp |= (hi as u64) << 32;
+    f64::from_bits(tmp)
+}
+
+#[inline]
+pub fn with_set_low_word(f: f64, lo: u32) -> f64 {
```
japaric: above comment applies here as well

src/math/mod.rs
japaric: Actually, this is fine given how it's actually used

PR #96, opened 2018-07-14
Implement atan and atanf

### LINKED ISSUE #8: implement atan
Source: https://git.musl-libc.org/cgit/musl/tree/src/math/atan.c

### LINKED ISSUE #42: implement atanf
Source: https://git.musl-libc.org/cgit/musl/tree/src/math/atanf.c

maxdeviant:
Implement `atan` and `atanf`.

Closes #8 
Closes #42

### DIFF
--- a/src/math/atan.rs
+++ b/src/math/atan.rs
@@ -1,0 +1,124 @@
+/* atan(x)
+ * Method
+ *   1. Reduce x to positive by atan(x) = -atan(-x).
+ *   2. According to the integer k=4t+0.25 chopped, t=x, the argument
+ *      is further reduced to one of the following intervals and the
+ *      arctangent of t is evaluated by the corresponding formula:
+ *
+ *      [0,7/16]      atan(x) = t-t^3*(a1+t^2*(a2+...(a10+t^2*a11)...)
+ *      [7/16,11/16]  atan(x) = atan(1/2) + atan( (t-0.5)/(1+t/2) )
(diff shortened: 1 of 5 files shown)


src/math/atan.rs
```diff
@@ -0,0 +1,124 @@
...
+
+#[inline]
+pub fn atan(x: f64) -> f64 {
+    let mut x = x;
+    let mut ix = x.to_bits();
+    let sign = ix >> 31;
+    ix &= 0x7fffffff;
+    if ix >= 0x44100000 {
+        /* if |x| >= 2^66 */
+        if x == f64::NAN {
```
maxdeviant: In `atanf` I used `isnanf(x)`, but there isn't a corresponding `isnan` for `f64`s.

Is checking against `f64::NAN` okay? And if so, should I use `== f32::NAN` instead of `isnanf`?

maxdeviant:
Having some trouble with `atan` failing the tests, so breaking into separate PRs.

See #98 for `atanf`.

src/math/atan.rs
P1n3appl3: Oh `isnanf()` was something japaric wrote because he didn't realize that `f32::is_nan()` was in core, not std. You should be able to do `x.is_nan()`.

crlf0710: um, NAN == NAN is false, right?

PR #99, opened 2018-07-14
Replace isnanf with f32::is_nan

### LINKED ISSUE #94: mod.rs needs an isnan for f64
I was just going to add it myself but have no idea how.

maxdeviant:
As discussed in #94 this PR removes `isnanf` in favor of `f32::is_nan`.

Closes #94.


japaric:
Thanks for the PR. This is already being done in #84

PR #106, opened 2018-07-14
implement fmaf

### LINKED ISSUE #20: implement fmaf
https://git.musl-libc.org/cgit/musl/tree/src/math/fmaf.c

erikdesjardins:
closes #20


japaric:
Interesting. For now I would say let's use [these dummy implementations of the fenv functions](https://git.musl-libc.org/cgit/musl/tree/src/fenv/fenv.c). Eventually we want to do the Rigth Thing for each target architecture but that will probably require inline assembly (`asm!`) and I'd like to not add unstable features right now. Let's make an issue about filling in the fenv stuff later.

PR #119, opened 2018-07-16
Implement `atan`

derekdreery:
I took the work by @maxdeviant and made the tests pass.

I've deviated a little from the musl function, using the INFINITY constants. Therefore I'd appreciate some eyes on this to make sure it's correct. There are some tests for standard values (30 degrees, 45 degrees etc.).

Any questions let me know!


src/math/atan.rs
```diff
@@ -0,0 +1,171 @@
...
+    let mut x = x;
+    let mut ix = (x.to_bits() >> 32) as u32;
+    let sign = ix >> 31;
+    ix &= 0x7fff_ffff;
+    if ix >= 0x4410_0000 {
+        if x.is_nan() {
+            return x;
+        }
+
+        return if sign != 0 { f64::NEG_INFINITY } else { f64::INFINITY };
```
japaric: This doesn't look correct. The original C code has:

``` c
z = atanhi[3] + 0x1p-120f;
return sign ? -z : z;
```

But `z` doesn't eval to infinity. `0x1p-120f` is 2 ^ -120 (or `0x3870000000000000` in its integer representation) so `z` is close to `atanhi[3]` (`1.57079632679489655800e+00`). You can write `0x1p-120f` like [this](https://github.com/japaric/libm/blob/master/src/math/acos.rs#L60).

src/math/atan.rs
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    #[test]
+    fn zero() {
+        assert_eq!(atan(0.0), 0.0);
+    }
+
+    #[test]
+    fn infinity() {
+        assert_eq!(atan(f64::INFINITY), f64::INFINITY);
```
japaric: This test is not correct `atan(inf)` evals to (around) `PI / 2` because `tan(PI / 2)` is [asymptotic to +inf](https://en.wikipedia.org/wiki/Trigonometric_functions#/media/File:Trigonometric_functions.svg).

src/math/atan.rs
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    #[test]
+    fn minus_infinity() {
+        assert_eq!(atan(f64::NEG_INFINITY), f64::NEG_INFINITY);
```
japaric: This also needs to be fixed `atan(-inf)` evals to (around) `-PI / 2`.

Also, these last four unit tests won't be required once #120 lands because the test-generator will generate them.

derekdreery:
Yeah I had a mind blank when looking at the graph - sorry! I'll fix.

derekdreery:
I've fixed up the mistakes you found (I'm not sure what was going through my mind). The other decision is whether to leave the tests or remove them. I'll leave that up to you, my opinion is that they don't do any harm, and so may as well remain.

japaric:
I don't think the infinity tests will pass but we'll see.

Thanks for the PR, @derekdreery.

bors r+

derekdreery:
I think it rejected because of the negative code review on the PR.

japaric:
Yeah, I can't find a way to approve this PR from the mobile UI I'm using. Could you open a new PR? I should be able to r+ that one.

derekdreery:
No worries, I'll squash as well.

PR #128, opened 2018-07-23
Add some tests for pow

rrbutani:
Not sure if these are comprehensive, but they cover all the edge cases identified in the original source.

Also, I use std (for some constants) in the tests; is this okay?


rrbutani:
Thanks; swapped std for core.

alexcrichton:
Sorry for the super long delay here @rrbutani but thanks for the PR!

PR #131, opened 2018-07-25
omit bounds check in release mode

### LINKED ISSUE #129: Some symbols reference libcore panicking functions
I was curious to give out the new libm-for-wasm bits a spin and got a [small example working](https://play.rust-lang.org/?gist=868682091e3b0ce8ecf89257b362582c&version=stable&mode=release&edition=2015), but to my surprise it imported a few functions from the environment:

```
  (import "env" "_ZN4core9panicking18panic_bounds_check17h5517321e0731ca2aE" (func $core::panicking::panic_bounds_check::h5517321e0731ca2a (type $t0)))
  (import "env" "_ZN4core9panicking5panic17he062e0bd3417b514E" (func $core::panicking::panic::he062e0bd3417b514 (type $t1)))
```

which isn't good for wasm!

It looks like these are mostly around the `rem_pio2_large` function (and family of functions from the looks of it)

japaric:
this eliminates panicking branches in the optimized version of the functions. We keep the bounds
checks when running the test suite to check that we never do an out of bounds access.

This commit also adds a "must link" test that ensures that future changes in our implementation
won't add panicking branches.

closes #129

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,9 @@ matrix:
     - env: TARGET=x86_64-unknown-linux-gnu
     - env: TARGET=cargo-fmt
       rust: beta
+      # no-panic link test
+    - env: TARGET=armv7-unknown-linux-gnueabihf
+      rust: nightly
 
 before_install: set -e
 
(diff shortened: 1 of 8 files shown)


japaric:
cc @alexcrichton

bors r+

PR #142, opened 2018-08-01
Change `sqrt` to use wrapping newtypes

mb64:
This changes the `sqrt` implementation to use the `Wrapping` newtype around the `u32`s to make it not panic on overflows.  This seems to be enough to make `sqrt` work with overflow checks enabled.

Helps #4.

### DIFF
--- a/src/math/sqrt.rs
+++ b/src/math/sqrt.rs
@@ -77,6 +77,7 @@
  */
 
 use core::f64;
+use core::num::Wrapping;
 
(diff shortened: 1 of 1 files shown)


little-arhat:
@japaric hi, any plans on merging this one and related PRs?

alexcrichton:
Sorry for the very long delay on merging this but looks like this still passes tests (yay!) so I'm going to merge

PR #143, opened 2018-08-01
Fix the atan family of functions behavior with bounds checks

mb64:
Similarly to #142, this changes `atan2` to use wrapping addition.  `atan`, `atanf`, and `atan2f` were already OK with bounds checks.


alexcrichton:
Again sorry for the very long delay on merging this also still passes tests so merging!

PR #144, opened 2018-08-02
Enable usage of int_to_from_bytes feature

eldruin:
When compiling the library, rustc complained that the feature int_to_from_bytes had to be enabled. Hence this PR.


burrbull:
It's no longer relevant.

PR #148, opened 2018-09-10
exclude CI files from appearing on crates.io

ignatenkobrain:


alexcrichton:
Thanks for the PR, but I'm going to close this since this isn't something we've got the bandwidth to maintain (pristine packages on crates.io)

PR #153, opened 2018-10-16
Fixed uint overflow in floorf for negative exponents

m1el:
`floorf` has a bug in extracting the exponent from bits, where intermediate calculation results are `u32`, but can go negative for negative exponents.

The test case is very simple: calling `libm::floorf(0.5)` in debug mode will panic with the following message:

`'attempt to subtract with overflow', <path>/libm/src/math/floorf.rs:6:13`

I took a quick look at other similar functions and couldn't find problems. It would be nice to run tests in debug mode to check if there are any other places that cause similar underflows.

### DIFF
--- a/src/math/floorf.rs
+++ b/src/math/floorf.rs
@@ -12,7 +12,7 @@ pub fn floorf(x: f32) -> f32 {
         }
     }
     let mut ui = x.to_bits();
-    let e = (((ui >> 23) & 0xff) - 0x7f) as i32;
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks for the PR @m1el and sorry for the delay! Are you still interested in landing this? If so mind including a test that exercises the overflow? Other than that I think this looks good to me

alexcrichton:
Ah I went ahead and pushed up a test for this mentioned in the OP, and looks like tests are still passing, so merging! Again sorry for the delay but thank you for the PR!

PR #155, opened 2019-01-21
fix overflow panic on floorf

jackmott:


jackmott:
closing this as I see there is already a PR for this.

PR #159, opened 2019-04-14
fix build and test

Yoopo:
- refactor exact_chunks to chunks_exact
- refactor to_bytes which as been deprecated
- remove qemu-arm-rt dependence which as been deprecated and does not compile
- refactor newlib test generation code
- extract common code in a template folder


Cargo.toml
```diff
@@ -26,3 +26,4 @@ members = [
 
 [dev-dependencies]
 shared = { path = "shared" }
+libc = { version = "0.2.14", default-features = false }
```
gnzlbg: This libc version is really old. Why isn't this just `0.2` ?

alexcrichton:
Sorry I don't personally have time to maintain this repository

Cargo.toml
Yoopo: No particular reason, I just followed what I could found online.
https://doc.rust-lang.org/1.30.0/unstable-book/language-features/lang-items.html#using-libc

alexcrichton:
Ok I need to apply some fixes in this repo and took another look at this repository. The test fixes were actually quite deep that needed to happen so I just merged those in and this shouldn't be necessary any more. I'm going to close this, but thanks for the PR and sorry for the delay!

PR #162, opened 2019-05-02
move into musl dir

burrbull:
Preparing for adding alternatives like newlib.


alexcrichton:
I'm a bit out of the loop wrt newlib/musl, mind giving me a bit of a brief overview of the intended course of action here? I'm not sure what the purpose of the `musl` feature is in that I feel like we'd probably want to always pick *some* implementation to give to users, and we could weigh newlib vs musl if we'd like?

burrbull:
https://github.com/rust-lang-nursery/libm/issues/138

alexcrichton:
FWIW I don't have a ton of time to help maintain this library, but I want to make sure it's not entirely forgotten since platforms rely on it. I'd personally appreciate answers to questions which are a bit more expansive if possible than a curt "here's an issue number" link.]

If the goal here is simply speed, why is a new directory of musl implementations needed? We're always only going to have one implementation of each function, and it could be annotated as to where it came from presumably?

mati865:
cc @japaric

PR #164, opened 2019-05-10
A lot of cosmetic changes

burrbull:


src/math/sqrt.rs
```diff
@@ -131,19 +131,19 @@ pub fn sqrt(x: f64) -> f64 {
...
-        ix1 = ix1 << i as usize;
+        ix1 <<= i as usize;
     }
     m -= 1023; /* unbias exponent */
-    ix0 = (ix0 & 0x000fffff) | 0x00100000;
+    ix0 = (ix0 & 0x_000f_ffff) | 0x_0010_0000;
     if (m & 1) == 1 {
         /* odd m, double x to make it even */
-        ix0 += ix0 + ((ix1 & sign) >> 31).0 as i32;
+        ix0 = ix0 + ix0 + ((ix1 & sign) >> 31).0 as i32;
```
vks: Why this change? This seems inconsistent with the change above (`ix1 = ix1 << i` to `ix1 <<= i`).

burrbull: I made this change because clippy warning.
I can revert and make clippy shut up.

Due to the fact that the variable is present on the right more than once, compiler can't optimize this code. Both variants are equivalent.
You can see ASM here.
https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=58c8753c3ee4954820234f64cd081db5

vks: Why does clippy warn in this case? I would expect the code to be equivalent, it is just a bit more readable to use `+=`.

For looking at ASM, I recommend [rust.godbolt.org](https://rust.godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(j:1,lang:rust,source:'pub+fn+inc(mut+a:+i32)+-%3E+i32+%7B%0A++++a+%2B%3D+1%3B%0A++++a%0A%7D%0A%0Apub+fn+foo(mut+a:+i32)+-%3E+i32+%7B%0A++++a+%2B%3D+a%2B1%3B%0A++++a%0A%7D%0A%0Apub+fn+bar(mut+a:+i32)+-%3E+i32+%7B%0A++++a+%3D+a%2Ba%2B1%3B%0A++++a%0A%7D'),l:'5',n:'0',o:'Rust+source+%231',t:'0')),k:50,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:nightly,filters:(b:'0',binary:'1',commentOnly:'0',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'1',trim:'1'),lang:rust,libs:!(),options:'',source:1),l:'5',n:'0',o:'rustc+nightly+(Editor+%231,+Compiler+%231)+Rust',t:'0')),k:50,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4).

burrbull: Ok.

alexcrichton:
This quantity of changes can't really be reviewed really. While we can be somewhat confident it's still correct due to CI, I'm personally wary to merge a change that just changes a ton of style in one commit with no message other than "cosmetic changes". I would expect at least a rationale for changes, a list of what changes were made, how they were made, etc. Ideally each change would be in its own separate commit too

src/math/sqrt.rs
JonathanDygert: The warning is because it looks like an improper refactor of `x = x + 1` into `x += x + 1` instead of `x += 1`.

JonathanDygert:
From some quick experimenting, there are 3060 total clippy warnings, 2866 are "unreadable_literal" and 147 are "many_single_char_names", only leaving 47 other warnings.

I'd imagine there's no good way to split up changing those 2866 literals in many commits, so if acceptable those should probably just be all together in one commit. Assuming `cargo fix` will correct those, it would be easy for a trusted contributor to submit that PR.

As for the single character names, they're so common in this type of code you might just want to set clippy to allow them.

The remaining 47 warnings would be much easier to go through.

alexcrichton:
Is there a reason this needs to be done for all files simultaneously? These are largely (as far as I understand) straight translations of C code to Rust. I'm not sure why we would want to update the style and or try to basically apply clippy style lints to C code. I think I don't relaly understand the motivation for this change.

burrbull:
Ok. I'll try to split it on several small PRs.

PR #165, opened 2019-05-13
rem_pio2: actually return medium value for x ~ 2pi/2

m1el:
There's a discrepancy between musl's rem_pio2 and libm's rem_pio2:
The calculation result for medium case is not used for x ~ pi/2..2pi/2


m1el:
Adding the `#[must_use]` attribute to scoped `medium` function would ensure that all calls are meaningful, but I think it's not worth doing.

alexcrichton:
Thanks! Can a test be added for this since it presumably wasn't caught by the existing tests?

m1el:
I'll try to find a test case for this, but it might be the case that this is an optimization for this specific range.

Although it should be obvious from the code that currently the return value of `medium` is discarded, and control flow doesn't follow musl's implementation.

m1el:
Added tests for this change, public interface `sin(f64)` is affected.

m1el:
The reason general tests didn't catch it: this difference only affects some values in range `[0x400921fb00000000..0x400921fbffffffff]` or `[3.141592025756836..3.1415939331054683]`, which has a rate of ~1e-10.

alexcrichton:
@bors: r+

Hm ok, perhaps the random generation for floats could be updated to have a low-probability path to generate values near pi and/or multiples of pi?

m1el:
@alexcrichton RNG that generates "interesting" numbers more often is definitely doable, that's a great idea. It should probably be implemented under a different issue/PR.

PR #169, opened 2019-05-17
Add newlib exp2f

### LINKED ISSUE #136: port newlib's exp2f
Source: https://sourceware.org/git/gitweb.cgi?p=newlib-cygwin.git;a=blob;f=newlib/libm/math/wf_exp2.c;hb=HEAD

What needs to be done:

Schultzer:
Hi guys,
I'm planning to deal we some of the "newlib issues" but I'm not sure in what direction we should go.

Alex mention in #162 
>we'd probably want to always pick some implementation to give to users and we could weigh newlib vs musl if we'd like?

Which I agree with.

Should we just annotate the function with `#[cfg(source = "musl")]` | `#[cfg(source = "newlib")]`?

fix #136

### DIFF
--- a/src/math/exp2f.rs
+++ b/src/math/exp2f.rs
@@ -1,136 +1,33 @@
-// origin: FreeBSD /usr/src/lib/msun/src/s_exp2f.c
-//-
-// Copyright (c) 2005 David Schultz <das@FreeBSD.ORG>
-// All rights reserved.
-//
-// Redistribution and use in source and binary forms, with or without
-// modification, are permitted provided that the following conditions
-// are met:
-// 1. Redistributions of source code must retain the above copyright
-//    notice, this list of conditions and the following disclaimer.
-// 2. Redistributions in binary form must reproduce the above copyright
-//    notice, this list of conditions and the following disclaimer in the
-//    documentation and/or other materials provided with the distribution.
-//
-// THIS SOFTWARE IS PROVIDED BY THE AUTHOR AND CONTRIBUTORS ``AS IS'' AND
-// ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
-// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
-// ARE DISCLAIMED.  IN NO EVENT SHALL THE AUTHOR OR CONTRIBUTORS BE LIABLE
-// FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
-// DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS
-// OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
-// HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
-// LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
-// OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
-// SUCH DAMAGE.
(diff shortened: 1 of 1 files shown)


alexcrichton:
Looks like [these two tests pass today](https://gist.github.com/alexcrichton/76743275534fb8d05ef2abc4b0fe8668) but fail on ARM? Also I'm not entirely certain, but what's the purpose of replacing the function?

Schultzer:
I think the rationale from japaric is that the musl version use f64 for some of the functions, which will yield slow performance on device that doesn't support f64.

The main goal for this PR is just to settle in which direction we should go. 

Relating to the failing test, I don't really know, one thing too note is it's not related to #168 because it happens in both debug and release mode

alexcrichton:
Ok, makes sense! In light of that would you be able to benchmark this and the previous implementation? I suspect the failing test is due to jn/jnf using this function transitively and its output is tweaked. They're probably not incorrect but the test harness at least needs some form of an update to avoid the failing tests.

Schultzer:
I'm pretty sure that the falling test is not related to this PR. it happens on master too after #168.

I'm not entirely sure about the benchmark suite, can we assume that if we make a benchmark suite that runs in docker on CI that it would have f64 native and emulation depending on the target?

I don't have any embedding device to benchmark on and would like to have it set up on CI but if we cannot simulate real world, then I see little point in using it as a prof of performance, but I agree that we should have a benchmark suite but I think it should be in a separate PR.

burrbull:
Some of ported newlib fns you can take here.    https://github.com/burrbull/libm/tree/master/src/math/newlib

Schultzer:
Thanks, it's great work, I had a branch going but I've read into some troubles with the test generator because some of the functions had changed signature like `pub(crate) fn rem_pio2f(x: f32) -> (i32, f64)` -> `pub fn rem_pio2f(x: f32) -> (i32, f32, f32)`

That's why I started this PR to get going.

What do you think should we keep the musl or go for newlib?

On a side note do you have any idea why the tests are failing on ARM?

alexcrichton:
Thanks for opening https://github.com/rust-lang-nursery/libm/issues/170 @Schultzer, it does indeed look unrelated! I think with a rebase and with another confirmation that CI is green this should be good to go, but I'm also still curious myself to learn more about moving to newlib and would prefer to have concrete data when changing existing implementations.

Schultzer:
I agree Alex, I would like to hold this PR back for a bit, until we have a benchmark suit.

Schultzer:
Benchmark from master 
```
test exp2f     ... bench:           5 ns/iter (+/- 0)
```
Benhmark from this pr
```
test exp2f     ... bench:           0 ns/iter (+/- 0)
```

burrbull:
> Thanks for opening #170 @Schultzer, it does indeed look unrelated! I think with a rebase and with another confirmation that CI is green this should be good to go, but I'm also still curious myself to learn more about moving to newlib and would prefer to have concrete data when changing existing implementations.

When this repository was created, the main argument in favor of the musl was the ease of porting.

Schultzer:
@alexcrichton I know we are benchmarking on CI but, what do you think?

gnzlbg:
Are you calling the function with the same inputs both times? Otherwise the comparison isn't very meaningful.

Schultzer:
Could we benchmark against all f32, using the same value for binary and trinary functions?

alexcrichton:
FWIW a benchmark of 0ns/iter means that it's probably a bad benchmark because it's being optimized out. While I agree that this is simpler I still think that more investigation is needed into the performance to see whether it's worth it

kpp:
Hey guys!
Is this PR is blocked because of lack of good benches? I can help.

alexcrichton:
Yes this and [other PRs](https://github.com/rust-lang-nursery/libm/pulls) are blocked on evaluating how to proceed, and the best metric for now is speed and there hasn't been much benchmarking yet to see how to proceed.

tgross35:
Sorry this PR has been around so long. We merged this crate into the https://github.com/rust-lang/compiler-builtins repo, so this wouldn't be able to merge as is; however, I think there are unfortunately also some licensing concerns that would prevent us using anything from newlib anyway.

We might have a solution for implementations that don't use 64-bit floats, more to come at https://github.com/rust-lang/compiler-builtins/issues/832.

Issue #170, opened 2019-05-17
Investigate jn and jnf randomly falling on CI


Schultzer:
Hi guys,

`jn` and `jnf` are randomly failing on ARM. 

On master [Build](https://dev.azure.com/rust-lang/libm/_build/results?buildId=146)
```
cargo test --features checked musl-reference-tests --target arm-unknown-linux-gnueabi --release
...
failures:

---- jn_matches_musl stdout ----
thread 'main' panicked at 'INPUT: [114, 4594974205335009568] EXPECTED: [1617394868955] ACTUAL 0.0', /target/arm-unknown-linux-gnueabi/release/build/libm-a8dc3a3490b5fe96/out/musl-tests.rs:108:17
note: Run with `RUST_BACKTRACE=1` environment variable to display a backtrace.


failures:
    jn_matches_musl
```

On #169 [Build](https://dev.azure.com/rust-lang/libm/_build/results?buildId=148)
```
cargo test --features checked musl-reference-tests --target arm-unknown-linux-gnueabi
...

failures:

---- jn_matches_musl stdout ----
thread 'main' panicked at 'INPUT: [136, 4602429132083530282] EXPECTED: [15924833] ACTUAL 0.0', /target/arm-unknown-linux-gnueabi/debug/build/libm-a6d3571cbabba738/out/musl-tests.rs:108:17
note: Run with `RUST_BACKTRACE=1` environment variable to display a backtrace.

---- jnf_matches_musl stdout ----
thread 'main' panicked at 'INPUT: [29, 1061546867] EXPECTED: [86] ACTUAL 0.0', /target/arm-unknown-linux-gnueabi/debug/build/libm-a6d3571cbabba738/out/musl-tests.rs:16:17


failures:
    jn_matches_musl
    jnf_matches_musl
```

## Questions

- Is it related to the use of `wrapping_add` and `wrapping_sup`?
- Are the algorithms in musl/newlib relying on overflows?

Schultzer:
Just to add why I believe this is some kind of random fluke. https://dev.azure.com/benjamin0103/benjamin/_build/results?buildId=9

burrbull:
You can dbg! these fns with INPUT on different targets and find difference.

Schultzer:
Thanks I fixed the test in `jnf` but it had a cascading effect so the root to this is deep but it might lead me in the right direction.

alexcrichton:
[...]

PR #172, opened 2019-05-26
Add newlib sqrtf

lzutao:
I ported sqrtf from https://github.com/eblot/newlib/blob/master/newlib/libm/math/ef_sqrt.c .

r? @japaric 

Blocked on #173


lzutao:
cc #170

alexcrichton:

alexcrichton:
Is this just a reorganization of what's already there at this point? If so, why is this PR needed?

lzutao:

PR #173, opened 2019-05-26
Add newlib fdlibm

lzutao:
Ported from https://github.com/eblot/newlib/blob/master/newlib/libm/common/fdlibm.h


alexcrichton:
Thanks for the PR! But what is this doing? There's no changes to the public interface in this PR?

Schultzer:
This PR contains functions needed for #174

alexcrichton:
I'm going to close this since I'd rather not merge a PR that adds a module that nothing uses, but if it's needed for other PRs then it can be included in other PRs.

PR #174, opened 2019-05-26
Add newlib fmodf

lzutao:
Ported from https://github.com/eblot/newlib/blob/master/newlib/libm/math/ef_fmod.c

Blocked on #173

r? @japaric


lzutao:
Build failed because of #170

alexcrichton:

lzutao:

PR #176, opened 2019-05-27
Add newlib copysignf

lzutao:
Ported from https://github.com/eblot/newlib/blob/master/newlib/libm/common/sf_copysign.c


alexcrichton:
Thanks for the PR! I'm not really sure why this is changing though? This is the same implementation as before?

tcyrus:
Seems like this was updated

Lokathor:
I'm with alex, where is the meaningful change?

lzutao:

PR #177, opened 2019-05-28
Add newlib cbrtf

### LINKED ISSUE #134: port newlib's cbrtf
Source: https://sourceware.org/git/gitweb.cgi?p=newlib-cygwin.git;a=blob;f=newlib/libm/common/sf_cbrt.c;hb=HEAD

What needs to be done:

lzutao:
Ported from https://sourceware.org/git/gitweb.cgi?p=newlib-cygwin.git;a=blob;f=newlib/libm/common/sf_cbrt.c;hb=HEAD

Block on #173 

Closes #134


lzutao:
I honestly don't know why it failed. And I need some review!

alexcrichton:

lzutao:
This implementation does not use `double` (`f64`) type. So it could run on
hardwares that have no support for double precision floats.

alexcrichton:
Ok that makes sense to me, although does Rust run on any architectures that are like that? (are there examples today of platforms that would benefit?) 

Review-wise there's no need to keep the musl code around, we only need one implementation of this function in-tree.

If the purpose here as well is to avoid `f64`, could some form of CI check be added to assert that `f64` isn't used?

Schultzer:
@lzutao can you please rebase your PR and align them with #169 so we only have one impl. in the tree? We got a benchmark suite so we can evaluate each function now.

lzutao:
I don't understand what you mean. This PR doesn't have any relations
with exp2f in #169 .

lzutao:
Ok. I rebased on master and add your commit in #169.

Schultzer:
@lzutao I'm sorry what I meant was that we only want one implementation in the tree, we don't want to have a musl module or a newlib module. The benchmark suite is to give a rough estimate, ideally this should be tested on hardware.

lzutao:
@Schultzer Do you know what wrong with this PR?
The ported cbrtf has the same generated asm with C version: https://godbolt.org/z/tfeV0c

Schultzer:
The errors here are expected, the system on the ci uses musl and, as far as I know, given the different implementation in newlib, we properly want to test this function against newlib, or increase the ULP for this case.

src/math/cbrtf.rs
```diff
@@ -12,65 +11,52 @@
...
-        /* zero or subnormal? */
-        if hx == 0 {
-            return x; /* cbrt(+-0) is itself */
-        }
-        ui = (x * x1p24).to_bits();
-        hx = ui & 0x7fffffff;
-        hx = hx / 3 + B2;
+    let mut t: f32 = if FLT_UWORD_IS_SUBNORMAL(hx) {
+        /* subnormal number */
+        let t: f32 = f32::from_bits(0x4b80_0000);
```
gnzlbg: This can be a const.

src/math/mod.rs
```diff
@@ -344,3 +344,97 @@ fn with_set_low_word(f: f64, lo: u32) -> f64 {
 fn combine_words(hi: u32, lo: u32) -> f64 {
     f64::from_bits((hi as u64) << 32 | lo as u64)
 }
+
+mod fdlibm {
+    /* @(#)fdlibm.h 5.1 93/09/24 */
```
gnzlbg: Can the items (fns and const) in this module be `pub(crate)` as opposed to just `pub` ?

lzutao: Done in #173 .

src/math/cbrtf.rs
lzutao: Done. But rustc would optimize it away as show in https://godbolt.org/z/tfeV0c.

lzutao:
> benchmarks say this is faster. I see no reasons against making this change.

I don't expect the ported newlib functions could be faster than old one. The reason
for newlib is to run on system have no hardware support for double precision floats.
See more in the issue #134 .

gnzlbg:
@lzutao so will this be a performance regression on systems that do have support for double precision floats ?

lzutao:
At least in x86_64, it doesn't have regression:
This PR:
```
test cbrtf     ... bench:          19 ns/iter (+/- 0)
```
Master:
```
test cbrtf     ... bench:          20 ns/iter (+/- 0)
```

alexcrichton:
It looks like this doesn't change the performance, so is this still necessary?

burrbull:
It should be tested on different platform, especially on embedded. With/without f32/f64 FPU.
Benchmark system should support different targets.

lzutao:

PR #185, opened 2019-06-19
Add benchmark suite

Schultzer:
Let me know what you guys think about this.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,6 +30,8 @@ members = [
 
 [dev-dependencies]
 no-panic = "0.1.8"
+rand = "0.6.5"
+paste = "0.1.5"
 
 [build-dependencies]
 rand = { version = "0.6.5", optional = true }
--- a/src/math/pow.rs
+++ b/src/math/pow.rs
@@ -479,7 +479,7 @@ mod tests {
             .for_each(|s| s.iter().for_each(|val| pow_test(base, *val, expected)));
     }
 
-    fn test_sets(sets: &[&[f64]], computed: &Fn(f64) -> f64, expected: &Fn(f64) -> f64) {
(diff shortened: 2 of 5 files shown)


alexcrichton:
Nice! This looks like a great start to me. Would it be possible to run this on CI in a dedicated job? That way we could keep everything else on stable and only test the benchmarks in one nightly location. I don't think we can get cross-architecture performance really, but we can certainly benchmark on Intel!

Schultzer:
I totally agree. I'll look into setting it up for as a dedicated job.

Schultzer:
@alexcrichton I've added a dedicated job and refactored the code.

Cargo.toml
```diff
@@ -30,6 +30,18 @@ members = [
...
 [dev-dependencies]
 no-panic = "0.1.8"
+rand = "0.6.5"
+paste = "0.1.5"
 
 [build-dependencies]
 rand = { version = "0.6.5", optional = true }
+
+[profile.bench]
+opt-level = 3
```
alexcrichton: Oh one thing I Forgot to mention is that this section can likely be almost entirely removed. The only setting which is set to something that's not the default is `lto = true`, but that probably isn't too useful for these benchmarks since it may not represent what callers will be benchmarking. (you typically don't LTO with libm)

src/math/pow.rs
```diff
@@ -479,6 +479,7 @@ mod tests {
             .for_each(|s| s.iter().for_each(|val| pow_test(base, *val, expected)));
     }
 
+    #[allow(bare_trait_objects)]
```
alexcrichton: Could this be updated with `dyn Fn`? Should be fine to allow and avoids a lint attribute!

Schultzer: done

Cargo.toml
Schultzer: done

Issue #150, opened 2018-10-02
Validate against actual libm implementations


gnzlbg:
The functions taking a single `f32` could be fully validated (for all input bit-patterns) against `libm`.

For functions taking multiple arguments and single `f64`s, the validation against `libm` could happen for some value ranges only.

Schultzer:
Hey @alexcrichton as far as I understand then the tests cover this right?

alexcrichton:
I think this issue is more about exhaustive testing of all f32 bit patterns whereas today we just do a best-effort randomized fuzzing sorta against musl's libm

tgross35:
We now have tests against MPFR with a handful of different generators, as well as benchmarks against musl, so I think this should be covered.

PR #186, opened 2019-06-21
Add test suite

Schultzer:
Hi guys.

I created a test suite. when running this in dev it spotted overflow in cos, fma, sin, sincos, tan functions.

let me know what you guys think.

I have not run the formatter yet.


alexcrichton:
Nice! I'm all for adding more tests, especially if they're just executed with simple `cargo test` commands :)

gnzlbg:
Looks good to me in general. 

I wonder whether we could auto-generate at least some of these tests, e.g., via a proc macro like:

```rust
#[libm_test(type = "trigonometric", include = "pi/2", ulp = "4.0")]
fn sind(x: f64) -> f64; 
#[libm_test(include(x = "2"))]
fn powf(x: f32, y: f32) -> f32; 
```

Many of these tests appear to be going for corner cases inputs, e.g., +-INFINITY, +-0, sqrt(...), pi / n, 2^n, e^n, etc. and their immediate neighborhoods. Depending on the type of function, some other regions are also worth testing, and well for binary inputs we can compute the cartesian product, and for unary f32 functions we can actually test all inputs. We could generate tests like this, and then use ULPs to compare the results against, e.g., the systems libm, requiring 0.5 ulps by default, and allowing the user to pass in a higher number as part of the proc macro, to allow for slight deviations in, e.g., transcendental functions.

For the manual tests, it might be interesting to consider moving the test inputs and results to some sort of external file, that's loaded by the test suite, so that, e.g., in case of a modification that introduces differences, one might just need to do something like `--bless` to update the testsuite, as opposed to having to go through all of these by hand.

@Schultzer I think adding these manual tests is good for now, but do you think these approach might be worth exploring in the future?

Schultzer:
@gnzlbg do you know of any projects they're doing something similar?
I've done this one time before with my blas impl. where I used JSON-FORTRAN to generate fortran result into a json file then having serde deserialize them. but I feel like including serde isn't really an option here.

I like your idea with the proc macro. I'm gonna look into that.

gnzlbg:
> do you know of any projects they're doing something similar?

Kind of. stdsimd auto-generates tests using a proc macro. 

>  but I feel like including serde isn't really an option here.

Why do you think that? AFAICT these dependencies only need to be dev-dependencies.

Schultzer:
My thinking was because one of the goals are to include this in core. But if there is no limit on dev tools then I’ll refactor this draft

alexcrichton:
FWIW we only include the library source in libcore, not the test dependencies. It's fine to pull in whatever is needed here for testing this crate.

Schultzer:
ping @alexcrichton and @gnzlbg, I've done the first part and moved the test out and implemented a proc-macro for the nearest function.
This was the easiest way I could find for making it more DRY.

I'm planning to implement one for each trigonometric, exponential, power, hyperbolic, error, gamma, and manipulation functions, etc.

Please let me know what you guys think about this approach.

alexcrichton:
Hm I'm not really sure I fully understand this, why are tests generated via a macro rather than being written as before? Is there enough shared between each set of tests that it's worth having in a macro?

gnzlbg:
@alexcrichton for each type of math functions you often want to test
similar input domains. For example, for the trigonometric functions you
often want to test all the floats close to 0, +-pi/4, pi/2, 3pi/4, etc. So
while for each of the functions one might want to have extra tests, testing
the classic bunch is something that one wants to do by default against all
of them.

alexcrichton:
While that's true the randomized testing we already do I think covers a good portion of that. I'm also not sure why it requires extraction into a full-blown procedural macro vs just writing test cases by hand.

gnzlbg:
> I'm also not sure why it requires extraction into a full-blown procedural macro vs just writing test cases by hand.

This doesn't require a proc macro, but I do think that hardcoding all inputs and outputs is a bad idea, except for super special cases. 

With a good testing crate, one can write this code easily by hand, e.g., 

```
fn sind(x: f64)->f64;

#[cfg(test)]
mod tests {
  #[test] fn sind_zero() { libm_test::trigonometric(|x| sind(x), Around(0_f64), Ulp(4)); }
  #[test] fn sind_pi_half() { libm_test::trigonometric(|x| sind(x), Around(PI / 2.), Ulp(4)); }
...
  #[test] fn sind_special() { libm_test::assert_system(|x| sind(x), Exact(MySpecialVal), Ulp(2)) }
}
```

With the proc macro we can just generate all the boiler plate tests and the code can focus on the special ones instead, e.g., 

```rust
#[libm_test(type = "trigonometric", ulp = "4")] 
fn sind(x: f64)->f64;

#[test] fn sind_special() { libm_test::check_output(|x| sind(x), Input(MySpecialVal), Ulp(2)) }
```

crates/libm-test/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "libm-test"
+version = "0.1.0"
+authors = ["Benjamin Schultzer <benjamin@schultzer.com>"]
+
+[lib]
+proc-macro = true
```
gnzlbg: You probably want to separate the library that does the tests from the proc macro that just generates tests and calls into that library.

src/math/floorf.rs
```diff
@@ -1,10 +1,13 @@
 use core::f32;
+#[cfg(test)]
+use libm_test::nearest;
 
 /// Floor (f64)
 ///
 /// Finds the nearest integer less than or equal to `x`.
 #[inline]
 #[cfg_attr(all(test, assert_no_panic), no_panic::no_panic)]
+#[cfg_attr(test, nearest)]
```
gnzlbg: All unary functions taking 32-bit arguments can just be tested against all inputs (all 32-bit bit patterns). I don't know why would it make sense to special case any tests here.

alexcrichton:
I primarily just want to make sure that tests are as readable and easy to modify as possible, as I think that's the priority rather than ensuring we have deduplicated testing and such. The cost of a macro for something like this I feel is pretty high, especially if the macro has a lot of complexity.

Schultzer:
Hey @alexcrichton, I've been working on this with @gnzlbg and decided that the first part will be to move the musl tests out of the build script and into the test-framework crate.

Let me know what you think.

src/math/acos.rs
```diff
@@ -62,6 +64,7 @@ fn r(z: f64) -> f64 {
 /// Returns values in radians, in the range of 0 to pi.
 #[inline]
 #[cfg_attr(all(test, assert_no_panic), no_panic::no_panic)]
+#[cfg_attr(test, libm_test)]
```
burrbull: `#[cfg_attr(test, libm_test::libm_test)]`

looks shorter then
```
#[cfg(test)]
use libm_test::libm_test;
#[cfg_attr(test, libm_test)]
```

gnzlbg:
@alexcrichton I haven't reviewed this yet. But to give you a more concrete idea of where this is going, the proc_macro attribute is intended to generate the tests for each function. This doesn't need to be a proc_macro per-se, and the current implementation is very "proc_macro"y, but we can change that. 

This macro, currently, generates the same random tests that were being generated before against musl, but now we test against the system's libm (which would be musl for musl targets, glibc for others, etc.). It also adds a couple more tests. For example, for unary `f32` functions, it tests all bit-patterns when compiled with `--release`. On CI, when running under qemu, this test is disabled, because it took too long. This type of testing has detected a lot of issues already. This is going to add a couple of more tests going forward:

* tests that check interesting ranges of values for each type of function, e.g., see how sleef does it here: https://github.com/shibatch/sleef/blob/master/src/libm-tester/tester3.c#L261 (it pookes at ~100000 bit patterns close to values where precision is known to behave poorly for the numerical algorithms)  - these will be enabled under qemu everywhere

* add mpfr as a dependency for testing, and test against the exact result. Some math functions are allowed by the standard to have a precision higher than 0.5 ULPs. Now, if the allowed error is 3 ULPs, musl has a 2 ULP error, and we have a 2 ULP error from musl, we are at 4 ULPs and 1 ULP too wrong. MPFR will give us a ground truth, and the system libm test will tell us "how wrong" are the system libraries in general with respect to that. This would also help if MPFR has bugs.

That would finish the testing part, at least for the time being. Afterwards, these proc macros will (probably) also be used to automatically generate benchmarks (not only randomly, but also for specific input domains that are know to take a longer time to compute, etc.). That should allow us to fix the precision issues while being aware of potential performance regressions.

crates/test-framework/src/conformance.rs
```diff
@@ -0,0 +1,699 @@
+use proc_macro2::{Ident, TokenStream};
+
+fn equalf64() -> TokenStream {
```
gnzlbg: Notice that this function does not need to be expanded all the time, it is just a normal function, that can be part of the test framework, and that the tests should call.

Everything that can belong in the test framework as "normal" (non quoted) Rust code should be. We should try to expand the minimum amount of Rust code to cut down compile-times.

crates/test-framework/src/conformance.rs
```diff
@@ -0,0 +1,699 @@
...
+        }
+        let ulps = (xi - yi).abs() as usize;
+        ulps <= max_ulp_diff
+      }
+    }
+}
+
+fn equalf32() -> TokenStream {
+    quote! {
+      fn equal(x: f32, y: f32, max_ulp_diff: usize) -> bool {
```
gnzlbg: Same here. Move this as a function into the test framework, that the tests just call.

crates/test-framework/src/conformance.rs
```diff
@@ -0,0 +1,699 @@
...
+          return false;
+        }
+        let ulps = (xi - yi).abs() as usize;
+        ulps <= max_ulp_diff
+      }
+    }
+}
+
+fn equali32() -> TokenStream {
+    quote! {
```
gnzlbg: Same as above.

crates/test-framework/src/conformance.rs
```diff
@@ -0,0 +1,699 @@
...
+    }
+}
+
+macro_rules! forward {
+  ($func:ident, $x:ident) => (
+      pub fn $func(ulp: usize) -> TokenStream  {
+        let equal: TokenStream = paste::expr!{ [<equal $x>]() };
+        quote! {
+          #[cfg(test)]
+          mod conformance {
```
gnzlbg: This is also expanded many times. Do you think it could be possible to move this into the test framework, and use generic parameters to avoid expanding this code for all functions ?

gnzlbg:
Some fixes have landed in master so this will need rebasing.

Schultzer:
This has been refactored, and includes the validation suite.

 closes #150

crates/libm-test/src/lib.rs
```diff
@@ -0,0 +1,65 @@
+extern crate core;
```
gnzlbg: I think you might be able to stamp a top-level `!#[cfg(test)]` here for the whole crate.

Schultzer:
@alexcrichton let me know if you want me to squash the commits.

crates/libm-test/Cargo.toml
```diff
@@ -0,0 +1,9 @@
+[package]
+name = "libm-test"
+version = "0.1.0"
+authors = ["Benjamin Schultzer <benjamin@schultzer.com>"]
+
+[dependencies]
+libm = { version = "0.*", path = "../../" }
```
alexcrichton: I think this may want to unconditionally enabled the `checked` feature perhaps? Or reexport it so we can test with/without the feature?

crates/libm-test/src/validation.rs
```diff
@@ -0,0 +1,118 @@
...
+}
+
+#[cfg(test)]
+extern "C" {
+    pub fn sincosf(x: f32, sin: *mut f32, cos: *mut f32);
+}
+
+#[test]
+pub fn sincosf_validation() {
+    if !cfg!(optmaized) {
```
alexcrichton: typo?

Schultzer:
It's been moved to libm-test crate, and refactored to include validation for x86_64, and cases for nan and infinities, as mention above we plan to include precision testing via mpfr, so keeping it all together seems like the right way.

crates/libm-test/src/validation.rs
Schultzer: Thanks.

crates/libm-test/Cargo.toml
Schultzer: I've reexported the feature, it still unclear to me how this feature works? and what part of the code base it affects.

burrbull:
When you say mpfr, you mean [rug](https://docs.rs/rug/1.4.0/rug/struct.Float.html)?

alexcrichton:
> It's been moved to libm-test crate

Hm ok I see that but it seems to use sort of odd-looking macros  now and has to be exhaustively tested? Previously it would automatically test new functions and they'd conform to a known set of signatures. I'd prefer to avoid new macro soup and duplication if possible.

Schultzer:
> Hm ok I see that but it seems to use sort of odd-looking macros now and has to be exhaustively tested? 

The only part we are exhaustively testing is unary f32 functions. The exhaustive tests did catch some bugs.

> it would automatically test new functions

Thats one reason why I liked the proc-macro idea, then it would be explicit what we are testing, which is also it now.

> I'd prefer to avoid new macro soup and duplication if possible

Which duplication are you referring too?

I would argue that it's easier to change the tests and understand now, than before.

Schultzer:
@burrbull It could be rug or another crate there is quite a few.

Schultzer:
I plan to add some validation for f64 too, this could properly catch issues like #165

gnzlbg:
> Previously it would automatically test new functions and they'd conform to a known set of signatures.

Yes. That was the main advantage of the proc macro approach. Now new functions need to be manually added to the test suite. That's a one line change, but needs to be done manually.

Schultzer:
@alexcrichton would you prefer to just write these tests out in each function?

> This has to do with libcore itself being built with panic=unwind, but why not just link to std and have it provide these?

does that mean that when we call `0f32.acos()` it makes a call to `extern "C" { pub fn acosf(x: f32) -> f32; }`

burrbull:
I made initial commit for support test `sleef` against `rug::Float` if someone is interested
https://github.com/burrbull/sleef-rs/commit/289e46e5e6c0fd7deffe21b564552d7ad11f6428

Results:
```
Max ULP = 0.703125 at -4.944768947135878e307, Average ULP = 0.25104109375	test f64::u10::test_sin ... ok

Max ULP = 1.53125 at -1.7301104562983707e308, Average ULP = 0.329521328125	test f64::u35::test_cos ... ok
```

I think something like this can be done for `libm`.

alexcrichton:
Sorry I really do not have a ton of time to maintain this crate, and this PR alone has over 40 (!) comments already for what I was hoping would be very simple, adding some more tests. This is a very important crate to rust-lang/rust and arbitrary rewrites of infrastructure need to be carefully done rather than simply proposed off-the-cuff.

I see testing this crate as falling into a few major categories:

* One is handwritten tests for each function. For example regression tests due to bugs that have been fixed with specific inputs/outputs. 
* Tests against a reference implementation (like musl). This is what the build script covers today. We need to guarantee that this is present for all functions that are defined in musl.
* Eventually tests like fuzz tests or some sort of more random/exhaustive checking could be performed, or something like that.

I don't know why the reference tests are being rewritten in this PR. I don't know why you can't already just write normal unit tests for each function. I'm not really sure why there's a big rewrite happening here vs simply adding new tests in appropriate modules.

tgross35:
@Schultzer sorry this PR has been open for so long, but I am going to close it now. We have pretty good test coverage at this point against both musl and MPFR, as well as reasonable generators.

Schultzer:
Thanks, great to see that this project finally got a great driver behind the wheel, it’s been a battle being a passenger on this bus.

tgross35:
Help is always welcome if you are interested! Just give me another month or so to get some things a bit more organized before jumping too deep into anything :)

PR #188, opened 2019-07-01
Fixed rounding to negative zero

m1el:
Rounding of negative values near zero should return negative zero.

Discovered thanks to test suite by @Schultzer

### DIFF
--- a/src/math/round.rs
+++ b/src/math/round.rs
@@ -36,7 +36,12 @@ pub fn round(mut x: f64) -> f64 {
     }
 }
 
-#[test]
-fn negative_zero() {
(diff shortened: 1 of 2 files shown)


src/math/round.rs
```diff
@@ -35,3 +35,8 @@ pub fn round(mut x: f64) -> f64 {
         y
     }
 }
+
+#[test]
```
gnzlbg: Would you mind putting this behind a `#[cfg(test)]` ? Same below.

m1el: Done.

gnzlbg:
Thanks!

PR #189, opened 2019-07-01
Fix incorrect f32<->f64 casting in j1f/y1f

m1el:
There's a mismatch in float casts between musl's and libm j1f/y1f implementation.

Discovered thanks to test suite by @Schultzer


gnzlbg:
Thanks!

PR #191, opened 2019-07-02
Fixed a few int overflows

m1el:
As a result of fuzzing this crate, I've found a few int overflows

rem_pio2 had incorrect u32/i32 operations
remquo has a straight up int overflow


m1el:
Tests seem to have failed for other reasons, could you please re-run them?

alexcrichton:
Awesome, thanks!

PR #192, opened 2019-07-02
[WIP] Provide a cdylib with the libm C ABI

gnzlbg:
This PR adds a new crate `libm-cdylib` that implements the `<math.h>` API and that is ABI compatible with the platforms libm implementation. 

Since libm is usually dynamically linked, this allows replacing the libm implementation of any program in the platform with the Rust implementation, e.g., using `LD_PRELOAD` and similar APIs. 

The library is tested by just linking it to C programs that use `<cmath.h>` and checking that the library is called and produce expected "unique" results that differ from the system libraries. The C programs are linked to the system library, and then the math library is dynamically overriden.

Right now this works properly on macos, I'll iterate on the linux part of this on CI.

### DIFF
--- a/crates/libm-cdylib/src/lib.rs
+++ b/crates/libm-cdylib/src/lib.rs
@@ -1,3 +1,10 @@
+#![cfg(
+    // The tests are only enabled on x86 32/64-bit linux/macos:
+    all(unstable_rust,
+        any(target_os = "linux", target_os = "macos"),
+        any(target_arch = "x86", target_arch = "x86_64")
+    )
+)]
 #![allow(dead_code)]
 #![cfg_attr(not(test), feature(core_intrinsics, lang_items))]
 #![cfg_attr(not(test), no_std)]
--- a/crates/libm-cdylib/src/lib.rs
+++ b/crates/libm-cdylib/src/lib.rs
@@ -1,0 +1,153 @@
+#![allow(dead_code)]
+#![cfg_attr(not(test), feature(core_intrinsics, lang_items))]
+#![cfg_attr(not(test), no_std)]
+
+#[path = "../../../src/math/mod.rs"]
+mod libm;
+
+#[macro_use]
+mod macros;
+
+#[cfg(test)]
+mod test_utils;
+
+#[cfg(not(test))]
+#[panic_handler]
+fn panic(_info: &core::panic::PanicInfo) -> ! {
+    unsafe { core::intrinsics::abort() }
+}
+
+#[cfg(not(test))]
+#[lang = "eh_personality"]
+extern "C" fn eh_personality() {}
+
+// All functions to be exported by the C ABI.
+// Includes a test input/output pair for testing.
+// The test output will be used to override the
+// result of the function, and the test input
+// is used to call the overriden function from C.
+// This is needed to make sure that we are linking
+// against this libm during testing, and not the
+// system's libm.
+//
+//
+// FIXME: missing symbols: _memcpy, _memset, etc.
+export! {
+    fn acos(x: f64) -> f64: (42.) -> 42.;
+    fn acosf(x: f32) -> f32: (42.) -> 42.;
+    fn acosh(x: f64) -> f64: (42.) -> 42.;
+    fn acoshf(x: f32) -> f32: (42.) -> 42.;
+    fn asin(x: f64) -> f64: (42.) -> 42.;
+    fn asinf(x: f32) -> f32: (42.) -> 42.;
+    fn asinh(x: f64) -> f64: (42.) -> 42.;
+    fn asinhf(x: f32) -> f32: (42.) -> 42.;
+    // fn atan(x: f64) -> f64: (42.) -> 42.;
+    fn atanf(x: f32) -> f32: (42.) -> 42.;
+    fn atanh(x: f64) -> f64: (42.) -> 42.;
+    fn atanhf(x: f32) -> f32: (42.) -> 42.;
+    fn cbrt(x: f64) -> f64: (42.) -> 42.;
+    fn cbrtf(x: f32) -> f32: (42.) -> 42.;
+    fn ceil(x: f64) -> f64: (42.) -> 42.;
+    fn ceilf(x: f32) -> f32: (42.) -> 42.;
+    fn copysign(x: f64, y: f64) -> f64: (42., 42.) -> 42.;
+    fn copysignf(x: f32, y: f32) -> f32: (42., 42.) -> 42.;
+    //fn cos(x: f64) -> f64: (42.) -> 42.;
+    //fn cosf(x: f32) -> f32: (42.) -> 42.;
+    fn cosh(x: f64) -> f64: (42.) -> 42.;
+    fn coshf(x: f32) -> f32: (42.) -> 42.;
+    fn erf(x: f64) -> f64: (42.) -> 42.;
+    fn erfc(x: f64) -> f64: (42.) -> 42.;
+    fn erff(x: f32) -> f32: (42.) -> 42.;
+    fn erfcf(x: f32) -> f32: (42.) -> 42.;
+    fn exp(x: f64) -> f64: (42.) -> 42.;
+    fn expf(x: f32) -> f32: (42.) -> 42.;
+    // FIXME: not in C:
+    // fn exp10(x: f64) -> f64: (42.) -> 42.;
+    // fn exp10f(x: f32) -> f32: (42.) -> 42.;
+    fn exp2(x: f64) -> f64: (42.) -> 42.;
+    fn exp2f(x: f32) -> f32: (42.) -> 42.;
+    fn expm1(x: f64) -> f64: (42.) -> 42.;
+    fn expm1f(x: f32) -> f32: (42.) -> 42.;
+    fn fabs(x: f64) -> f64: (42.) -> 42.;
+    fn fabsf(x: f32) -> f32: (42.) -> 42.;
+    fn fdim(x: f64, y: f64) -> f64: (42., 42.) -> 42.;
+    fn fdimf(x: f32, y: f32) -> f32: (42., 42.) -> 42.;
+    fn floor(x: f64) -> f64: (42.) -> 42.;
+    fn floorf(x: f32) -> f32: (42.) -> 42.;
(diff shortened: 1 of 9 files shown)


crates/libm-cdylib/src/lib.rs
```diff
@@ -0,0 +1,153 @@
...
+#[macro_use]
+mod macros;
+
+#[cfg(test)]
+mod test_utils;
+
+#[cfg(not(test))]
+#[panic_handler]
+fn panic(_info: &core::panic::PanicInfo) -> ! {
+    unsafe { core::intrinsics::abort() }
```
gnzlbg: TODO: I should probably call the C abort function here.

crates/libm-cdylib/src/lib.rs
```diff
@@ -0,0 +1,153 @@
...
+}
+
+#[cfg(not(test))]
+#[lang = "eh_personality"]
+extern "C" fn eh_personality() {}
```
gnzlbg: There has to be a better way that avoids this, but I exhausted all my google-fu trying to find it.

crates/libm-cdylib/src/macros.rs
```diff
@@ -0,0 +1,55 @@
+macro_rules! export {
```
gnzlbg: TODO: document this.

crates/libm-cdylib/src/macros.rs
```diff
@@ -0,0 +1,55 @@
...
+            }
+            #[cfg(not(link_test))] {
+                libm::$id($($arg),*)
+            }
+        }
+
+        #[cfg(test)]
+        paste::item! {
+            #[test]
+            fn [<$id _test>]() {
```
gnzlbg: TODO: document this

crates/libm-cdylib/src/test_utils.rs
```diff
@@ -0,0 +1,130 @@
...
+    );
+    s
+}
+
+/// For a given Rust type `x`, this prints the name of the type in C,
+/// as well as the printf format specifier used to print values of that type.
+pub(crate) fn ctype_and_printf_format_specifier(x: &str) -> (&str, &str) {
+    match x {
+        "f32" => ("float", "%f"),
+        "f64" => ("double", "%f"),
```
m1el: should this be `%lf`?

gnzlbg: IIRC, %f prints doubles. There is no format string for printing floats AFAIK - %f converts floats into a double, and prints that.

Schultzer: Would it make sense when `%f` is a double anyway? maybe `%.24f` for floats `%.54f` for doubles makes more sense?

Schultzer: Or even better maybe just `e`

gnzlbg: For the linking tests it doesn't really matter. I'll add a comment about this.

gnzlbg: So I've added a comment explaining what this does right now. 

If we wanted to verify the results of this lib, I think I would prefer doing so as part of the libm-test system. We could dynamically link libm as the system library, and just verify that the results of the C ABI match the results that are produced by the Rust library, or something like that.

The problem which checking the ABI here is that, if for some reason linking failed, or the symbol has a slightly different name, etc. the functions from the system library are invoked. So if we were to just simply verify the results, we can't tell whether its our libm or the system libm that's being invoked.

gnzlbg: Maybe there is a way to reliably prevent libm from being linked on Linux, while still being able to link, e.g., `fprintf` and other parts of the standard library. On MacOSX, this is IMO not worth it - everything links `libSystem.dylib`, which always dynamically links `libm`.

Schultzer:
Is there anyway we can fix the linking issue?

By looking the at the online docs the following functions should be present in the Kernel
- atan
- cos, cosf
- j0, j1, jn

There are no sincos or exp10 functions in the Kernel, they are only present in the accelerate framework  as vectorized version.

gnzlbg:
The problem is that some symbols are missing from the cdylib, like `_memcpy`, `_memset`, etc. I don't know why, and would prefer to merge this as is, and try to fix each one of those issues in follow up PRs. This issue appears related: https://github.com/rust-lang/rust/issues/44871

cc @alexcrichton

crates/libm-cdylib/src/lib.rs
alexcrichton: This has to do with libcore itself being built with panic=unwind, but why not just link to `std` and have it provide these?

crates/libm-cdylib/src/lib.rs
```diff
@@ -0,0 +1,156 @@
+#![allow(dead_code)]
+#![cfg_attr(not(test), feature(core_intrinsics, lang_items))]
+#![cfg_attr(not(test), no_std)]
+
+#[path = "../../../src/math/mod.rs"]
+mod libm;
```
alexcrichton: Could this do `extern crate libm` instead?

crates/libm-cdylib/src/lib.rs
```diff
@@ -0,0 +1,156 @@
...
+    // FIXME: missing symbols - fails to link
+    // fn tan(x: f64) -> f64: (42.) -> 42.;
+    // fn tanf(x: f32) -> f32: (42.) -> 42.;
+    fn tanh(x: f64) -> f64: (42.) -> 42.;
+    fn tanhf(x: f32) -> f32: (42.) -> 42.;
+    // FIXME: missing symbols - fails to link
+    // fn tgamma(x: f64) -> f64: (42.) -> 42.;
+    // fn tgammaf(x: f32) -> f32: (42.) -> 42.;
+    fn trunc(x: f64) -> f64: (42.) -> 42.;
+    fn truncf(x: f32) -> f32: (42.) -> 42.;
```
alexcrichton: If we really want to maintain a cdylib in this repository then I would prefer that this list were automatically inferred instead of duplicated with the definitions in the main source code. If the duplication is going to happen here then I would prefer that this cdylib is provided in a different repository than this one.

crates/libm-cdylib/src/macros.rs
```diff
@@ -0,0 +1,72 @@
+macro_rules! export {
+    (fn $id:ident ($($arg:ident : $arg_ty:ty),* ) -> $ret_ty:ty:
+       ($($test_arg:expr),*) -> $test_ret:expr;) => {
+        #[no_mangle]
+        pub extern "C" fn $id($($arg: $arg_ty),*) -> $ret_ty {
+            // This just forwards the call to the appropriate function of the
+            // libm crate.
+            #[cfg(not(link_test))] {
```
alexcrichton: Adding a pretty complicated macro to test what already has a pretty complicated of tests seems like overkill? Why does the cdylib itself need to be tested when it's all already tested elsewhere?

crates/libm-cdylib/src/test_utils.rs
```diff
@@ -0,0 +1,143 @@
+use std::{fs, io, path::Path, process};
```
alexcrichton: I personally feel like this file may be a bit over-the-top when testing, I'm not sure it's really all that necessary to be invoking Cargo from this crate's own test suite to test various C things? 

If it's really this extensive then I would prefer this not live in this repository to help keep this repository manageable and not pull in a whole set of new maintenance.

crates/libm-cdylib/src/lib.rs
gnzlbg: That would pull in std, see below.

crates/libm-cdylib/src/lib.rs
gnzlbg: Because libstd would pull the system's libm.

crates/libm-cdylib/src/lib.rs
gnzlbg: Will do.

crates/libm-cdylib/src/macros.rs
gnzlbg: This tests that the C ABI we provide is compatible with `<math.h>`, which is what allows this library to be used from C. 

We currently don't test it, and in fact, `libm` currently doesn't provide a compatible ABI, which is why some of the functions in the `lib.rs` are commented out. 

Simplest example is the `libm` functions returning tuples by value (e.g. `(i32, f32)`). The libm C ABI doesn't do that, and takes pointer arguments instead that are used as "out" parameters.

crates/libm-cdylib/src/test_utils.rs
gnzlbg: It isn't expensive on my 8 year old system - `cargo` is invoked on every test, but once the first invocation completes, all of them just return "success - the library was already built". 

I can use a `lazy_static!` to just call `cargo` once, but it felt like unnecessary complexity.

crates/libm-cdylib/src/lib.rs
gnzlbg: I'm still going to have to generate some of these by hand though because of the API mismatches between this libm, and C's libm.

Schultzer: Yeah even the original test harness does that, but it's not as explicit why it does it, I had to look up the musl code to understand what was going on in the test harness.

gnzlbg:
So I've cleaned this up, and CI is green. I've added two MacOSX build jobs, and a nightly x86_64-linux build job. For the time being, the tests are only run there. 

This is sub-optimal. Ideally, we would have a `libm` library, that exposes an `extern "C"` API, and a `libm-rs` library, that wraps that API in the `F32Ext`/`F64Ext` traits. We would then only exhaustively test `libm` and only have minimal tests in `libm-rs`. When compiling `libm` as a `cdylib`, we would use `panic=abort`, while we compiling it as a `rlib`, we would use `panic=unwind` (to allow linking it against both `abort`/`unwind` code). 

But these are architectural decision that have to be made depending on the goals of the library, so I've opened #195 to discuss that first.

alexcrichton:
It seems like this is intended to build on https://github.com/rust-lang-nursery/libm/pull/198 so I'm going to hold off on reviewing/merging until that's landed.

PR #193, opened 2019-07-02
Add signum

Schultzer:

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -73,8 +73,7 @@ pub trait F32Ext: private::Sealed + Sized {
 
     fn abs(self) -> Self;
 
-    // NOTE depends on unstable intrinsics::copysignf32
-    // fn signum(self) -> Self;
+    fn signum(self) -> Self;
 
     fn mul_add(self, a: Self, b: Self) -> Self;
 
@@ -176,6 +175,15 @@ impl F32Ext for f32 {
     #[inline]
     fn abs(self) -> Self {
         fabsf(self)
+    }
+
+    #[inline]
+    fn signum(self) -> Self {
+        if self.is_nan() {
+            f32::NAN
+        } else {
+            copysignf(1., self)
+        }
     }
 
     #[inline]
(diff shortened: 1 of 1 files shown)


alexcrichton:
Hm this may have been a bit too quick to merge? This doesn't have any tests and the definition differs from all other definitions which have their own dedicated file.

Schultzer:
As far as I know `signum` does not exist in libm, this seams to be a rust specific function that utilize `copysign`. as we already test copysign, then we can properly just add doc tests like the ones in std.

gnzlbg:
@alexcrichton sorry, my fault - I should have argued why I was merging this so that you could review it quicker, since I actually had the same questions that you are having.

So I went fishing for these APIs in musl's lib, and like @Schultzer says, these functions are not there. These methods are however available in `libstd` for the `f32` and `f64` types (e.g. see [here](https://github.com/rust-lang/rust/blob/master/src/libstd/f32.rs#L187)). IIUC, the purpose of the F32Ext and F64Ext traits is to provide `no_std` crates with the same API as libstd, and this PR is one step closer towards that. Which is why I merged it. 

Let me know if this clarifies things. I'll try to leave a better review comment next time.  

---

I think it would be helpful to have this type of information in the PRs message / commit message as well (e.g. with links to libstd, or musl), so that reviewers don't have to go fishing for this information. Not that I am consistent in doing this properly myself (e.g. once you are the one implementing something, e.g., because the methods were missing, that they have to be there feels kind of obvious - but that might not be the case for others)., but I think we should strive towards that to keep standards high.

alexcrichton:
I don't think it's really too too worthwhile to create a mirror API to libstd/libcore in this crate and then try to promote this crate as an alternative. Rather we should be working to move APIs from libstd to libcore over time, and the APIs in libcore can be backed by this crate. I would personally prefer to avoid stopgap situations like this where it makes it less clear what the long-term prospects for this crate are.

gnzlbg:
We should remove the `F32Ext` and `F64Ext` traits from this crate then.

alexcrichton:
Yes those have been around for quite some time and I would personally advocate for their removal as well

gnzlbg:
So I agree, the document I posted in #195 covers this (these traits are problematics for other reasons like generating a cdylib).

PR #194, opened 2019-07-02
[do not merge] Add test case for #170

Schultzer:


Schultzer:
cc @m1el It does look like the issue in #170 has been fix right?

PR #196, opened 2019-07-03
Add remainder

Schultzer:
This PR adds the missing `remainder` and `remainderf` found in musl libm respectly https://git.musl-libc.org/cgit/musl/tree/src/math/remainder.c and https://git.musl-libc.org/cgit/musl/tree/src/math/remainderf.c


alexcrichton:
I think the interface here doesn't match musl? I think that may also mean there's a bug in the auto-generated tests since they didn't catch this?

Schultzer:
Great catch actually what we want to return here is just the float from `remque`.

Issue #195, opened 2019-07-03
Goals and usecases for libm


gnzlbg:
There are a bunch of changes being proposed and merged for libm that deal with the "architecture" and I have the feeling that everyone has a different idea of the goals of the project. 

I think it would be worth it to collect all possible goals and use cases, before we start discussing which ones of them do we want to pursue, and which ones would be out-of-scope. 
[I've written a small document that collects the goals that came to my mind.](https://paper.dropbox.com/doc/Goals-for-Rust-libm--AgOlAUNFtFswIE9Y8_BIhzgsAg-gRlnRke09h8fNcIeMGk4v) It can be collaboratively edited, so please do add everything that comes to your mind. 

Once that is done, I'll summarize it, and we can start having a meaningful discussion about which goals of that list should the project have. Once we have consensus on that, it will be easier to know on a case-by-case basis whether the changes being proposed are progress towards those goals or not, and which architecture we should have for achieving those goals.

(If necessary we could have a small 15 min meeting about the goals after the list is collected and the summary is written with stakeholders)

cc @alexcrichton @Schultzer @burrbull @m1el @rrbutani @rust-lang-nursery/libm

alexcrichton:
As of a few months ago this was basically a dead crate and/or on life support and we needed changes for wasm. I've tried to keep it alive since then and put it on a solid foundation again by getting testing up and running. I am not personally available to lead the charge on large refactorings, major new features, etc. 

That being said this is an extremely important crate now to the standard library and rust-lang/rust due to its *default* usage on basically all targets for some math intrinsics. This isn't a crate that should be taken lightly and pulled in whichever direction is necessary.

I don't think there's really an option right now to add as many features as possible here, but rather this crate needs to remain focused on being a streamlined implementation for various math functions that are commonly found in libm C libraries.

gnzlbg:
> That being said this is an extremely important crate now to the standard library and rust-lang/rust due to its default usage on basically all targets for some math intrinsics. 

Can you elaborate on how the standard library uses this crate ? I don't find any references to it in rust-lang/rust .

jacobrosenthal:
Related is if libm should be usable across the no_std ffi barrier. Quite
often you want to ffi some existing c code which needs libm, and libm needs
libc, but libc is generally conflicting so you dont want to use either if
possible. Its possible to manually copy paste and extern no mangle the libm
function I need, but this isn't particularly maintainable.
[...]

PR #198, opened 2019-07-04
Refactor testing framework into a libm-test

gnzlbg:
cc @alexcrichton 

This PR removes the old testing framework, and instead adds two new crates, libm-analyze, and libm-test. 

libm-analize provides a proc macro, that expands its argument for each of the public functions in the libm crate, providing it its signature:

```rust
macro_rules! user_defined_macro {
    (
        id: $id:ident;      // The function name
        api_kind: $api_kind:ident;  // The variant name in ApiKind enum for this fn
        arg_tys: $($arg_tys:ty),*;  // The types of function arguments
        arg_ids: $($arg_ids:ident),*;  // The names of function arguments
        ret_ty: $ret_ty:ty;  // The return type
    ) => {};
}
libm_analyze::for_each_api!(user_defined_macro);
```

This is then re-used by the `libm-test` crate, to implement the random tests for each of the libm APIs that were previously provided by the build-script. It uses the same proc macro to implement exhaustiveness tests ala #186 , and it replaces with it most of libm-bench as well. 

The `libm_analyize::for_each_api` macro does a sanity check of the libm API when the libm-analyze tests run. It marks dozens and dozens of issues (missing inline, missing no_panic, no extern "C", returning types that are not repr(C), mismatches with musl API, etc.). The list is in this [gist](https://gist.github.com/gnzlbg/06893e29c14692fd9ff869a7bf210315). Fixing all of those takes time, so the errors are silenced by default, except when building the libm-analyze crate. At some point we should turn these into hard errors. EDIT: I started doing that here, but I've left hrad API breaking changes to subsequent PRs.


crates/libm-test/tests/system_libm.rs
```diff
@@ -0,0 +1,215 @@
...
+                    return false;
+                }
+                if x.is_nan() && y.is_nan() {
+                    return true;
+                }
+                if x.is_infinite() != y.is_infinite() {
+                    // one is inf but the other is not
+                    return false;
+                }
+                if x.is_infinite() != y.is_infinite() {
```
Schultzer: This appear twice.

crates/libm-test/tests/exhaustive32.rs
```diff
@@ -0,0 +1,82 @@
...
+    ) => {
+        #[test]
+        #[allow(unused)]
+        fn $id() {
+            extern "C" {
+                // The system's libm function:
+                fn $id($arg_id: f32) -> $ret_ty;
+            }
+
+            for i in 0..u32::max_value() {
```
Schultzer: Would it be an idea to be inclusive here? 
```rust
for i in 0..=u32::max_value() {
```

crates/libm-test/tests/system_libm.rs
```diff
@@ -0,0 +1,135 @@
...
+                    *[$($e),*].choose(r).unwrap()
+                } else {
+                    r.gen::<$id>()
+                }
+            }
+        }
+    }
+}
+
+// Some interesting random values
```
Schultzer: We should properly think about adding positive and negative NAN, MIN_POSITIVE, MAX and, subnorm which are `1.401298464324817070923730e-45f32` and `4.9406564584124654e-324f64`

gnzlbg: I tried to keep the tests testing exactly the same thing as in the old `build.rs`, to keep the test from breaking on CI due to too many new issues - there were already enough simpler to fix issues being raised by libm-analyze. I think we should move this to do something cleverer here as we fix the tests.

crates/libm/src/math/sincosf.rs
```diff
@@ -23,7 +23,9 @@ const S2PIO2: f32 = 2.0 * PI_2; /* 0x400921FB, 0x54442D18 */
 const S3PIO2: f32 = 3.0 * PI_2; /* 0x4012D97C, 0x7F3321D2 */
 const S4PIO2: f32 = 4.0 * PI_2; /* 0x401921FB, 0x54442D18 */
 
-pub fn sincosf(x: f32) -> (f32, f32) {
+#[inline]
+#[cfg_attr(all(test, assert_no_panic), no_panic::no_panic)]
+pub extern "C" fn sincosf(x: f32) -> (f32, f32) {
```
burrbull: Here should be wrapper for compatibility with C. Maybe something like this:
```
#[inline]
pub extern "C" fn sincosf(x: f32, sin: &mut f32, cos: &mut f32) {
     let (s, c) = sin_cosf(x);
     *sin = s;
     *cos = c;
}

#[inline]
pub fn sin_cosf(x: f32) -> (f32, f32) {
 ...
}
```

gnzlbg: The main point of `extern "C"` is making the function `#[no_unwind]`.

gnzlbg: Why do you need Rust compatible wrappers for these APIs ?

gnzlbg: Ahhhh! @burrbull sorry! I understood your argument wrong! 

You mean, that the signature of these functions does not match the C one, right ?

If that's the case, yes, I agree that this should be fixed, but I haven't gotten around to it. The rust-analyze script reports that this is the case for the following APIs: 

```
[error]: Function "modf" returns unsupported type -> ( f64 , f64 )
[error]: Function "remquo" returns unsupported type -> ( f64 , i32 )
[error]: Function "frexpf" returns unsupported type -> ( f32 , i32 )
[error]: Function "remquof" returns unsupported type -> ( f32 , i32 )
[error]: Function "lgammaf_r" returns unsupported type -> ( f32 , i32 )
[error]: Function "sincosf" returns unsupported type -> ( f32 , f32 )
[error]: Function "modff" returns unsupported type -> ( f32 , f32 )
[error]: Function "lgamma_r" returns unsupported type -> ( f64 , i32 )
[error]: Function "sincos" returns unsupported type -> ( f64 , f64 )
[error]: Function "frexp" returns unsupported type -> ( f64 , i32 )
[error]: Function "modf" returns unsupported type -> ( f64 , f64 )
[error]: Function "remquo" returns unsupported type -> ( f64 , i32 )
[error]: Function "frexpf" returns unsupported type -> ( f32 , i32 )
[error]: Function "remquof" returns unsupported type -> ( f32 , i32 )
[error]: Function "lgammaf_r" returns unsupported type -> ( f32 , i32 )
[error]: Function "sincosf" returns unsupported type -> ( f32 , f32 )
[error]: Function "modff" returns unsupported type -> ( f32 , f32 )
[error]: Function "lgamma_r" returns unsupported type -> ( f64 , i32 )
[error]: Function "sincos" returns unsupported type -> ( f64 , f64 )
[error]: Function "frexp" returns unsupported type -> ( f64 , i32 )
```

burrbull: I really noob in FFI. What signature will this function have for C user?
```
void sincos(double x, double *sin, double *cos);
```

gnzlbg:
The FMA implementation is quiet broken. Instead of using `wrapping_..` appropriately, it uses the non-wrapping operators everywhere. I've disabled the system tests for it, even though I landed a couple of bug fixes. 

I've removed nexttoward/nexttowardf for similar reasons. I don't think we can implement these correctly in Rust without a f80 floating point type with 64-bit significand.

crates/libm/src/math/sincosf.rs
danielhenrymantilla: Can the pointers in the C API be NULL? Just something to be aware of before exposing to C a function that UBs on NULL input.

burrbull: http://git.musl-libc.org/cgit/musl/tree/src/math/sincos.c#n16

danielhenrymantilla: Good, the pointers cannot indeed be NULL, however, given the C example, instead of `&mut f32`, `&Cell<f32>` should be used.

gnzlbg: A cell is not `f32` pointer sized. Maybe we need `NonNull<T>` here ?

danielhenrymantilla: The closest equivalent to C's `integer_type *` (a nullable mutable pointer to a potentially aliased integer), is `Option<&Cell<integer_type>>`.
The `Option` can be skipped because C implementation does not check against that,  and given that this function is not thread-safe anyways, the `&Cell` could be replaced by a `&mut`, I guess. One last question, though: is `f32` the portable equivalent of `C`'s `double` ?

gnzlbg: > The closest equivalent to C's numeric_type * (a nullable mutable pointer to a potentially aliased integer), is Option<&Cell<numeric_type>>.

Using a raw pointer:

```rust
let sin = MaybeUninit::uninit();
let cos = MaybeUninit::uninit();
sincos(x, sin.get_mut_ptr(), cos.get_mut_ptr()); 
```

is ok, because a raw pointer *mut T can point to an invalid T - since the T is uninitialized, the T is, in general, invalid. (*)

Using a `&Cell<T>`  , `&T` requires `T` to be _valid_. Since `Cell<T>` stores a _valid_ `T` behind a `UnsafeCell<T>`, and `UnsafeCell<T>` requires its contents to be valid, one cannot use `&Cell<T>` to initialize a memory location for an invalid `T`.

Using `NonNull<T>` this is not an issue either, since `NonNull<T>` stores a `*mut T`, which can point to an invalid `T`. 

cc @ralfjung - I kind of wish `MaybeUninit` would return `NonNull<T>`, so that we can use `NonNull<T>` in these APIs. 

---


(*) the only `T`s that would be valid here are those for which the "all bits are uninitialized" is a valid bit-pattern - but right now this is not specified for any Ts.

gnzlbg: > One last question, though: is f32 the portable equivalent of C's double ?

No: `f32` and `f64` are not guaranteed to be layout compatible with C's `float` and `double` types. This should use `libc::c_float` and `lib::c_double`, which are guaranteed to be layout compatible with C's `float` and `double` types. There is no Rust type compatible with C's `long double`, nor can we in general support C's `float_t` and `double_t` types either, due to a lack of a `f80` type.

gnzlbg: Since this is kind of out-of-scope for this PR, I've opened an issue to discuss this: https://github.com/rust-lang-nursery/libm/issues/199 .

crates/libm-test/tests/system_libm.rs
gnzlbg: I've added some of these values, haven't added the subnormal yet, I thought `MIN_POSITIVE` and `-MIN_POSITIVE` where the two subnormal values. Is that incorrect ?

alexcrichton:
I'm sorry I just can't keep reviewing multi-thousand line complete reorganizations of crates. Can this be split up? Some changes I would prefer to not happen after a skim are:

* The main sources of the crate should continue to live in `/src/*.rs`, not in `/crates/libm/src/*.rs`
* Rust-defined functions should not have `extern` ABIs, only if we add a C ABI crate should they be added.

I again do not want to lose the current functionality of tests which I'm afraid a refactoring like this could possibly lose. Can all these changes be disentangled from one another to avoid having a massive blocking PR?

alexcrichton:
Also if things like the fma implementation are broken can a separate PR be used to track that? It gets very difficult to manage this when so many fixes are piled in one place.

Additionally I do not want to add clippy to CI here, it seems like it will cause endless headaches of maintenance over time as clippy lints change.

gnzlbg:
> I again do not want to lose the current functionality of tests which I'm afraid a refactoring like this could possibly lose.

This change does not lose any tests AFAICT, all APIs are automatically tested, and compilation fails if one is not. One can ignore the results of tests of some APIs, if these happen to be broken, but all original tests pass here.

> Rust-defined functions should not have extern ABIs, only if we add a C ABI crate should they be added.

That's being discussed in #195 so I'd rather wait to see how that resolves there.

> Can this be split up?

Sure.

gnzlbg:
@alexcrichton so PRs #203, #204, #205, #206, #207, #208, and #209 are the ground work for this. After those get resolved. I'll rebase here and see what the next steps are. 

The first step would be to just submit a PR adding the `libm-analyze` crate, and migrating the existing tests to use it. Once that is resolved, further PRs would then add in parallel the remaining tests, and migrate libm-bench to use it as well.

alexcrichton:
Ok, thanks!

PR #202, opened 2019-07-08
Add missing no-panic

Schultzer:


Schultzer:
I wonder if there is function we always want to panic?

alexcrichton:
This seems good to add, but looks like some bugs need to be fixed and/or no_panic implementations commented out temporarily.

Schultzer:
I'm closing this, in favor of #207

PR #204, opened 2019-07-09
Move unit tests into its own file

gnzlbg:
This PR moves all unit tests into their own file so that they are not included when pulling the math module via a `#[path]` import (like the one in compiler builtins, or in a crate that wants to build a cdylib).


alexcrichton:
I'm not sure what the reason for this is? I personally think it's great to be able to write tests inline and have them executed. You're in general not really supposed to use `#[path]` to import crates but even if it were used these are always configured out, right?

gnzlbg:
The compiler-builtin crate uses `#[path]`, and since people forget to `#[cfg(test)]` their tests out, this means that `#[test]` end up always compiled into compiler-builtins unconditionally.

gnzlbg:
Alternatively, we could properly `cfg()` all tests here, but then we should have a way to enforce that no `#[test]` end up being generated for compiler builtins.

In the long term, I would prefer these tests to move to the `libm-test` crate, since they are going to require more tools to actually work (the `_eq` functions at the top level of the libm crate do not cut it).

gnzlbg:
The last commit moves the musl reference tests and required functionality into its own file.

alexcrichton:
It may be worthwhile to point out that `#[test]` implicitly expands to `#[cfg(test)]`. There is no need to use both `#[cfg(test)]` and `#[test]` on a function.

I continue to not think that this change is worth it since it's more readable to have tests inline and I don't think there's a reason to move them out.

PR #206, opened 2019-07-09
Properly test wasm32-unknown-unknown on CI

gnzlbg:
Currently not all configurations are build for wasm32, and the llvm intrinsics are incorrectly feature gated.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,8 +29,12 @@ members = [
   "crates/libm-bench",
 ]
 
+[build-dependencies]
+rand = { version = "0.6.5", optional = true }
+
 [dev-dependencies]
 no-panic = "0.1.8"
 
-[build-dependencies]
-rand = { version = "0.6.5", optional = true }
+[target.wasm32-unknown-unknown.dev-dependencies]
+wasm-bindgen-test = "0.2.47"
+
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -39,17 +39,9 @@ jobs:
           TARGET: powerpc64le-unknown-linux-gnu
         x86_64:
           TARGET: x86_64-unknown-linux-gnu
-
-  - job: wasm
-    pool:
-      vmImage: ubuntu-16.04
-    steps:
-      - template: ci/azure-install-rust.yml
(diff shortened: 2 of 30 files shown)


alexcrichton:
I don't think this is something we want to do on CI? Wasm is different enough that we don't really want to test all configurations exhaustively, we just want to make sure that it builds at all and doesn't hit codegen errors. The CI errors here for wasm are sort of just the start I'd imagine for getting the full suite running.

gnzlbg:
So this is ready for now. Depending on whether #204 is merged or not, what would remain doing is using the `wasm_bindgen_test` macro. I find it would be easier to do when all tests are in the same file.

gnzlbg:
So I've added use of `wasm_bindgen_test` here. As you can see, I had to add a `#![cfg(not(test))]` to the `compiler_builtin_smoke` test to avoid including the tests. 

I can workaround that here if you want, but the more tests we add directly into the libm crate, the more workarounds are going to be required in the `compiler_builtin_smoke` crate, and in the `compiler-builtin`s crate, to either duplicate the testing framework of libm so that `compiler_builtin`s can be tested, or to silence including the tests there. 

I still think that, since this crate is only used by including it via `#[path]`, the cleanest solution is to have `libm` be just the code that gets included, and to move all the tests to a different crate, to avoid having to add workarounds for this all over the place. In the mean time, `#![cfg(not(test))]` works for the compiler builtins smoke test, and would also be a one line change for the compiler-builtin test, although disabling all tests there might not be desired. 

I suppose we could add `--cfg`s here that only enable stuff when testing, but from the experience with #198 , we are going to need quiet a few of them (for exhaustiveness tests, mpfr tests, etc.).

gnzlbg:
Note: the failures here are unrelated and due to https://github.com/rust-lang/rust/pull/62574 , should be fixed in nightly soon.

alexcrichton:
I'm personally not really super keen on merging this unfortunately. This seems to add a good deal of implementation complexity and infrastructure, which is just yet-again-more to keep working on CI. I don't think we really get any benefit from running these tests on wasm either, so I'm not really sure it's worth the weight of this PR to merge it?

PR #209, opened 2019-07-10
Fix overflow bugs in fma

gnzlbg:
This fixes some of the overflow bugs in fma. It probably has others.


alexcrichton:
Can this use the current style of the library which is to include all tests inline with the module itself?

gnzlbg:
Done

tests/unit.rs
```diff
@@ -0,0 +1,16 @@
+use libm::*;
```
alexcrichton: Stray file now perhaps?

PR #210, opened 2019-07-10
Remove most `#[inline]` annotations

alexcrichton:
These annotations fall into a few categories

* Some simply aren't needed since functions will always be in the same
  CGU anyway and are already candidates for inlining.
* Many are on massive functions which shouldn't be inlined across crates
  due to code size concerns.
* Others aren't necessary since calls to this crate are rarely inlined
  anyway (since it's lowered through LLVM).

If this crate is called directly and inlining is needed then LTO can
always be turned on, otherwise this will benefit downstream consumers by
avoiding re-codegen'ing so many functions.

### DIFF
--- a/src/math/acos.rs
+++ b/src/math/acos.rs
@@ -48,7 +48,6 @@ const QS2: f64 = 2.02094576023350569471e+00; /* 0x40002AE5, 0x9C598AC8 */
 const QS3: f64 = -6.88283971605453293030e-01; /* 0xBFE6066C, 0x1B8D0159 */
 const QS4: f64 = 7.70381505559019352791e-02; /* 0x3FB3B8C5, 0xB12E9282 */
 
-#[inline]
 fn r(z: f64) -> f64 {
     let p: f64 = z * (PS0 + z * (PS1 + z * (PS2 + z * (PS3 + z * (PS4 + z * PS5)))));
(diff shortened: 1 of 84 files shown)


gnzlbg:
LGTM. These are only compiled once in compiler-builtins, where the symbols for these are generated. I don't know how much does `#[inline]` affect the ability of LLVM to do LTO. 

Fixing the current benchmarks to always use the same values, most benchmarks don't change that much. There are some that get much slower, and some that get a bit faster. See https://gist.github.com/gnzlbg/b282db61d37e305b7ca153d41f179f7c (new is this PR, old is master).

I'd rather add inline when it's needed, than everywhere, but the current benchmarks are kind of useless for that.

gnzlbg:
Thinking about this some more, I think I would prefer for the remaining `no_panic` checks to be added _before_ landing this.

alexcrichton:
A regression of 95ns/iter isn't really what's happening though:

```
 sqrt       0            95                     95     inf%   x 0.00 
 sqrtf      0            26                     26     inf%   x 0.00 
```

that basically means it was constant-folded before so we weren't actually benchmarking anything.

I'm going to go ahead and merge this since the tests are green, and we can continue to add `no_panic` annotations, or abandon `no_panic` entirely if it's not really measuring the thing we want.

PR #216, opened 2019-08-07
Improve sqrt/sqrtf if stable intrinsics allow

Lokathor:
Per the question in https://github.com/rust-lang-nursery/libm/issues/214, this uses the intel intrinsics if compile time settings allow.

### DIFF
--- a/src/math/sqrt.rs
+++ b/src/math/sqrt.rs
@@ -95,146 +95,146 @@ pub fn sqrt(x: f64) -> f64 {
             }
         }
     }
-    #[cfg(target_feature="sse2")]
+    #[cfg(target_feature = "sse2")]
     {
-      // Note(Lokathor): If compile time settings allow, we just use SSE2, since
-      // the sqrt in `std` on these platforms also compiles down to an SSE2
-      // instruction.
-      #[cfg(target_arch="x86")]
-      use core::arch::x86::*;
-      #[cfg(target_arch="x86_64")]
-      use core::arch::x86_64::*;
-      unsafe {
-        let m = _mm_set_sd(x);
-        let m_sqrt = _mm_sqrt_pd(m);
-        _mm_cvtsd_f64(m_sqrt)
-      }
+        // Note(Lokathor): If compile time settings allow, we just use SSE2, since
+        // the sqrt in `std` on these platforms also compiles down to an SSE2
+        // instruction.
+        #[cfg(target_arch = "x86")]
+        use core::arch::x86::*;
+        #[cfg(target_arch = "x86_64")]
+        use core::arch::x86_64::*;
+        unsafe {
+            let m = _mm_set_sd(x);
+            let m_sqrt = _mm_sqrt_pd(m);
+            _mm_cvtsd_f64(m_sqrt)
+        }
     }
-    #[cfg(not(target_feature="sse2"))]
+    #[cfg(not(target_feature = "sse2"))]
     {
-      let mut z: f64;
-      let sign: Wrapping<u32> = Wrapping(0x80000000);
-      let mut ix0: i32;
-      let mut s0: i32;
-      let mut q: i32;
-      let mut m: i32;
-      let mut t: i32;
-      let mut i: i32;
-      let mut r: Wrapping<u32>;
-      let mut t1: Wrapping<u32>;
-      let mut s1: Wrapping<u32>;
-      let mut ix1: Wrapping<u32>;
-      let mut q1: Wrapping<u32>;
+        let mut z: f64;
+        let sign: Wrapping<u32> = Wrapping(0x80000000);
+        let mut ix0: i32;
+        let mut s0: i32;
+        let mut q: i32;
+        let mut m: i32;
+        let mut t: i32;
+        let mut i: i32;
+        let mut r: Wrapping<u32>;
+        let mut t1: Wrapping<u32>;
+        let mut s1: Wrapping<u32>;
+        let mut ix1: Wrapping<u32>;
+        let mut q1: Wrapping<u32>;
 
-      ix0 = (x.to_bits() >> 32) as i32;
-      ix1 = Wrapping(x.to_bits() as u32);
+        ix0 = (x.to_bits() >> 32) as i32;
+        ix1 = Wrapping(x.to_bits() as u32);
 
-      /* take care of Inf and NaN */
-      if (ix0 & 0x7ff00000) == 0x7ff00000 {
-          return x * x + x; /* sqrt(NaN)=NaN, sqrt(+inf)=+inf, sqrt(-inf)=sNaN */
-      }
-      /* take care of zero */
-      if ix0 <= 0 {
-          if ((ix0 & !(sign.0 as i32)) | ix1.0 as i32) == 0 {
-              return x; /* sqrt(+-0) = +-0 */
-          }
-          if ix0 < 0 {
-              return (x - x) / (x - x); /* sqrt(-ve) = sNaN */
-          }
-      }
-      /* normalize x */
-      m = ix0 >> 20;
-      if m == 0 {
-          /* subnormal x */
-          while ix0 == 0 {
-              m -= 21;
-              ix0 |= (ix1 >> 11).0 as i32;
-              ix1 <<= 21;
-          }
-          i = 0;
-          while (ix0 & 0x00100000) == 0 {
-              i += 1;
-              ix0 <<= 1;
-          }
-          m -= i - 1;
-          ix0 |= (ix1 >> (32 - i) as usize).0 as i32;
-          ix1 = ix1 << i as usize;
-      }
-      m -= 1023; /* unbias exponent */
-      ix0 = (ix0 & 0x000fffff) | 0x00100000;
-      if (m & 1) == 1 {
-          /* odd m, double x to make it even */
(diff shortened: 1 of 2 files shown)


Lokathor:
I don't know what the process of picking a reviewer is, other than that I'm pretty sure I shouldn't be reviewing my own PRs of course. GitHub has a suggested reviewer, but also gives their status as "busy". I'm not sure who else to ping for this, but I guess everyone subscribed to the repo gets an email so whoever will see this.

* The PR is ready to be reviewed and merged.
* This is a "patch" level change, the public API is unchanged.

src/math/sqrt.rs
```diff
@@ -95,128 +92,150 @@ pub fn sqrt(x: f64) -> f64 {
...
+        // instruction.
+        #[cfg(target_arch = "x86")]
+        use core::arch::x86::*;
+        #[cfg(target_arch = "x86_64")]
+        use core::arch::x86_64::*;
+        unsafe {
+            let m = _mm_set_sd(x);
+            let m_sqrt = _mm_sqrt_pd(m);
+            _mm_cvtsd_f64(m_sqrt)
+        }
```
gnzlbg: Doesn't using the same code here as for wasm32 produce the exact same effect for targets with SSE2? If not, why not ?

Lokathor: I think I don't understand the question. `sse2` is an x86 and x86_64 feature. It will never be available during a wasm32 build.

gnzlbg: I meant that instead of using the `core::arch` intrinsics here, you can just write `unsafe { ::core::intrinsics::sqrtf64(x) }` just like wasm32 does, and that should generate this exact same code if SSE2 is enabled.

gnzlbg: For example, see: https://rust.godbolt.org/z/eEF9ef

Lokathor:
we can add cfg-if and use that, but it's not a current dependency and i didn't want to add one carelessly.

we can certainly add some more tests.

i think that refactoring the whole wasm32 path vs every other path thing _in general_, which happens in almost every single function, is a little out of scope for this PR. We should definitely have a bigger discussion about that separately. I'd hope that's not a blocker to this particular PR.

gnzlbg:
> i think that refactoring the whole wasm32 path vs every other path thing in general, which happens in almost every single function, is a little out of scope for this PR. 

FWIW I wasn't suggesting that, only suggesting that for the functions for which more than the 2 current paths are exposed, adding more `#[cfg()] + #[cfg(not(...))]` paths starts getting out-of-hand.

> we can add cfg-if and use that, but it's not a current dependency and i didn't want to add one carelessly.

@alexcrichton should decide on that, we can always c&p the macro here somewhere if that's better than adding a dependency.

src/math/sqrt.rs
Lokathor: ah, but that's not stable. this code is. that's why i did not also add a neon version, I'm only out to help out stable rust today.

you're right that we should probably also reconsider  calling the core intrinsic in more cases if Nightly is in use.

src/math/sqrtf.rs
```diff
@@ -29,83 +27,103 @@ pub fn sqrtf(x: f32) -> f32 {
...
-    let mut ix: i32;
-    let mut s: i32;
-    let mut q: i32;
-    let mut m: i32;
-    let mut t: i32;
-    let mut i: i32;
-    let mut r: u32;
+    #[cfg(target_feature = "sse")]
+    {
+        // Note(Lokathor): If compile time settings allow, we just use SSE, since
```
alexcrichton: I'd remove the "Lokathor" part of the note here, it's not really directed at any one person per se.

Lokathor: oh, sure, it was "note from lokathor", in case people had questions later, but i can leave the name off in the future.

gnzlbg:
> This is not true, each intrinsic has hundreds of randomized tests generated to compare against musl upstream

Whether this is true or not depends on what those 500 random tests are doing. When I checked, some of the intrinsics had less than 30% code coverage no matter how often I regenerated those tests. So sure, the library does have some tests, but I don't think these tests are, by far, sufficient to test a new implementation of any math function.

Schultzer:
AFAIK This libm impl. is the only one out there that do random tests, all other use the same test vector but test for precision and spec. I would argue that random test is not really something you want to rely on, they are great as early indicator.

Lokathor:
SO, here's the deal: @gnzlbg told me last night [on zulip](https://rust-lang.zulipchat.com/#narrow/stream/122651-general/topic/libm) that _apparently_ I've been using libm wrong this whole time.

* This PR is based around the idea that you'd have to declare a libm dependency, call the function, and then rustc builds the crate and all that.
* It turns out that because libm is special you can just declare an `extern "C"` block with all the functions you want and the functions _will be there_ (in a "non-guaranteed", "not our fault if it breaks" sort of way) because either the platform's C libm will have them or compiler-builtins will use Rust libm to provide them.
* If you do depend on and then build libm yourself you get worse code in many situations because llvm knows about libm and will often enough optimize your code for you (eg: `sqrt(0.0)` gets const folded to  just `0.0`).

BASED ON THIS NEW INFO: I'm going to just close up this particular PR for now, no changes needed. When I get time I'll petition T-libs to just make these functions stable in core since "they're already there" anyway.

gnzlbg:
@Lokathor You _can_ use `libm` by just adding it as a dependency. There are different ways of using `libm` that might be better for your use case, depending on what you are doing. 

The best way to use these is to use `libstd`, but since you are kind of building your own, you probably can't do that. 

Using the symbols is a possibility, but you need to make sure that the symbols exist, and you probably won't get any LLVM optimizations, like the one you have implemented in this PR. The only "advantage" of using the symbols over libm is that if your target has a C math library, those functions will be called instead, but that's pretty much it. 

If you want to get LLVM optimizations like libstd does, but in a `#![no_std]` library, you either need to use `core::intrinsics` or `link_llvm_intrinsics` to call the right LLVM intrinsics, and make sure that the symbols are available. If after optimizations the symbol is called, and that resolves to a libm function, you kind of want those to be optimized properly. In this particular case, for these particular SSE2 targets, LLVM never calls the symbol, but this isn't true in general, so it makes sense to optimize the functions here (it is just unclear whether it is worth it to provide stable Rust code that does that instead of a 1-line of LLVM code when we know that LLVM does the right thing, e.g., like for wasm32).

If you don't care about using the math functions from a C library and LLVM optimizations for the math functions, or want to use the exact same math functions for multiple targets independently of what the C library does, or many other use cases, using libm directly might be the best choice.

---

Relying on the symbols existing is super-brittle, and will probably break if we ever allow any kind of user customization for these symbol names like GCC and Clang allow. Currently, `libcore` has some tests that rely on some of the symbols existing. (e.g. [see here]((https://github.com/rust-lang/rust/blob/master/src/libcore/tests/num/flt2dec/mod.rs#L88))), so if you are considering relying on that, you might want to, first run libcore tests for the targets you care about.

gnzlbg:
> If you don't care about using the math functions from a C library or about LLVM optimizations, or want to use the exact same math functions for multiple targets independently of what the C library does, or are constrained to stable Rust, or many other use cases, using libm directly might be the best choice. 

For those use cases, this PR does improve things, so it might be worth it to keep it open or merge it. It is however unclear whether the use cases that would be satisfied by this aren't better satisfied by exposing the math functions from libcore.

Lokathor:
Ah, hmm, if it's still an improvement then I'll open it back up.

src/math/sqrtf.rs
Lokathor: Comments updated

Lokathor:
Alright,

* fixed the comments that were mentioned in review.
* Opened a new issue for discussion of code path selection improvement https://github.com/rust-lang-nursery/libm/issues/217

Ready for re-review and possible merge if it passes review.

alexcrichton:
Yes this won't actually ever be used on x86/x86_64 targets that have sse with libstd because `1.0.sqrt()` will cause LLVM to do the equivalent of what this is already doing, but @Lokathor you're thinking that you'd still like this to land for any users either explicitly using symbols and/or this library itself? Do you know of users doing that on x86/x86_64?

Lokathor:
well, me for one.

More broadly, anyone else who depends on libm as a Stable-rust, core-only source of math operations will need this update.

Particularly, some gamedev-wg folks and I have been trying to improve the state of linear algebra libraries for graphics, and in several benchmarks nalgebra had 10x or worse time taken compared to other common libraries because the `std` feature flag had been left off the build which caused `nalgebra` to switch to a pure-software sqrt _even on_ x86/x86_64 where they could have stayed with a stable hardware sqrt given enough conditional compilation. We can do that tricky work on behalf of others and then tell them "just use libm, it does the right thing".

I agree with gnzlbg that the correct _long term_ path is to work on getting more core::intrinsics stabilized, but the PR for that RFC hasn't seen any activity in 2019 at all ([reminder link](https://github.com/rust-lang/rfcs/issues/2505)), and there's a rustc [tracking issue](https://github.com/rust-lang/rust/issues/50145) for it that's had some success but also seems to not be moving along very quickly. Given this, I think it's appropriate to make this improvement here.

alexcrichton:
Ok that sounds reasonable to me!

Before we fill this out for other intrinsics I'd prefer to get the macro story and style sorted out, but moving forward with at least this one sounds reasonable to me.

PR #218, opened 2019-08-10
Fix broken link in README

kpp:


Lokathor:
Hi kpp! This is a good catch, thanks for the fix.

Unfortunately, I'm the newest maintainer here. I'm going to approve this, but we'll wait on someone else to make sure that the correct merge procedure is followed, if any. They only put me on the team last Tuesday, so I haven't approved and merged anything yet.

PR #221, opened 2019-08-14
slightly improve spec and sanity check coverage

Lokathor:
Per the general desire to have better tests in the crate, this is just a small touch up to the functions that have been of interest lately.

Doing too many functions at once makes it hard to review, so we're keeping it to just eight functions: f32 and f64 versions of `sqrt`, `ceil`, `floor`, and `abs`.


alexcrichton:
👍

PR #224, opened 2019-09-05
Refresh the CI setup

Lokathor:
This is its own PR _just_ to fix the CI/features, we can do the other thing with sqrt separately once we're confident we have this part fully sorted out.

* Remove the `checked` feature, just use the `debug_assertions` flag that's already part of Rust
* Swap the `stable` feature when we want Stable to be an `unstable` feature when we want Nightly bonuses.
* There are no longer any default features. The crate should operate in a totally plain mode by default.
* Since `#![deny(warnings)]` has been in the `lib.rs` the whole time, and since my editor runs clippy on the source, I at least silenced all the clippy warnings that we definitely don't care about. There's a lot more to do on the clippy front, but that means doing little tiny changes to almost every single file, so we'll leave that out for now

---

I'm putting up this PR so people can look and I can start to see where the CI errors are, I really don't expect it to pass CI this first time through.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,17 +11,15 @@ version = "0.2.0"
 edition = "2018"
 
 [features]
-# only used to run our test suite
-default = ['stable']
-stable = []
+default = []
+
+# This tells the compiler to assume that a Nightly toolchain is being used and
+# that it should activate any useful Nightly things accordingly.
+unstable = []
 
 # Generate tests which are random inputs and the outputs are calculated with
 # musl libc.
 musl-reference-tests = ['rand']
-
-# Used checked array indexing instead of unchecked array indexing in this
-# library.
-checked = []
 
 [workspace]
 members = [
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -49,8 +49,6 @@ jobs:
         displayName: "Install rust wasm target"
       - script: cargo build --target wasm32-unknown-unknown
         displayName: "Build for wasm"
-      - script: cargo build --target wasm32-unknown-unknown --no-default-features
-        displayName: "Build for wasm (no default features)"
     variables:
       TOOLCHAIN: nightly
 
--- a/ci/run.sh
+++ b/ci/run.sh
@@ -3,7 +3,7 @@
 set -ex
 TARGET=$1
 
-CMD="cargo test --all --no-default-features --target $TARGET"
+CMD="cargo test --all --target $TARGET"
 
 $CMD
 $CMD --release
--- a/ci/run.sh
+++ b/ci/run.sh
@@ -5,11 +5,14 @@ TARGET=$1
 
 CMD="cargo test --all --target $TARGET"
 
+# stable by default
 $CMD
 $CMD --release
 
-$CMD --features 'stable'
-$CMD --release --features 'stable'
+# unstable with a feature
+$CMD --features 'unstable'
+$CMD --release --features 'unstable'
 
-$CMD --features 'stable checked musl-reference-tests'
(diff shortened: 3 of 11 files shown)


alexcrichton:
Looks reasonable to me!

Lokathor:
* The PowerPC build seems to have failed because of a crates.io fluke, because 0.6.5 is a valid version of rand and all, <https://crates.io/crates/rand/0.6.5>
* The ARM build seems to be something to do with the `jn` function, but as far as I can tell that file wasn't affected by any change in the PR so far so I don't know why it hit a failure.

Can anyone else tell what's wrong with `jn`?

gnzlbg:
> Can anyone else tell what's wrong with jn?

Apparently it gives the wrong answer:

> 'INPUT: [140, 4603707074953425158] EXPECTED: [24658263201] ACTUAL 0.0'

You might want to open a bug. I'll retry this, that input is generated randomly.

alexcrichton:
There's a [known issue with `jn` and `jnf`](https://github.com/rust-lang-nursery/libm/issues/170) which would explain that.

Lokathor:
Alright, then I'm marking this as ready for review and merge

alexcrichton:
Oh right one last thing before merging, @Lokathor would you be up for updating the submodule in rust-lang-nursery/compiler-builtins and getting CI passing there as well?

Lokathor:
Uh, I guess? I'm not familiar with how to make it run CI against an un-released version.

Which reminds me that this needs to be a breaking version bump for this libe since it's a big breaking change.

Lokathor:
oh silly me, 0.2.0 isn't released on crates.io yet so this can all still be 0.2.0

Lokathor:
Oh it's a **git** submodule. I thought it was like a Cargo thing.

I'm afraid I can't help with that. Every time I do anything more than just commit and push things go badly for me. I don't know how git submodules work at all.

alexcrichton:
Yes I basically mean just send a PR to the compiler-builtins repo and get CI green there. You can update the submodule by editing `.gitmodules` to your git repository and then going into the folder and resetting the git repo to the revision of this PR.

Lokathor:
i can at least give it a try later today

Lokathor:
... so I tried to use `git reset` on a clone of the repository but it didn't appear to do anything

```
D:\dev\compiler-builtins>git reset

D:\dev\compiler-builtins>cd libm

D:\dev\compiler-builtins\libm>git reset

D:\dev\compiler-builtins\libm>
```

(Unfortunately I genuinely don't know how `git` works at all, I just use the Windows GUI thing and that's the limit of my `git` skills.)

I think whatever git magic you want tested has to be done by someone else.

alexcrichton:
A git submodule is a url and a sha for a commit, so you can update the URL by editing `.gitmodules` and then you need to check the actual submodule to a particular commit to commit it within the parent repository. That means you'll need to go into the `libm` submodule, add a remote necessary to fetch your branch, reset it to this commit, and then commit everything. That probably looks something like:

```
$ cd compiler-builtins
# edit .gitmodules
$ cd libm
$ git remote add lokathor https://github.com/Lokathor/libm
$ git fetch lokathor
$ git reset --hard lokathor/new-CI
$ cd ..
$ git add libm .gitmodules
$ git commit -m "Update the libm submodule"
```

Lokathor:
The current file has no commit hash in it, https://github.com/rust-lang-nursery/compiler-builtins/blob/master/.gitmodules,

But I did the other steps and opened the PR anyway, <https://github.com/rust-lang-nursery/compiler-builtins/pull/313>

alexcrichton:
Ok looks like the compiler-builtins PR is green, so let's merge!

PR #226, opened 2019-09-10
Migrate from azure pipelines to Github actions

alexcrichton:
Should make user management easier and also helps follow the repository
when it's renamed.


.github/workflows/main.yml
```diff
@@ -0,0 +1,67 @@
...
+        - mips-unknown-linux-gnu
+        - mips64-unknown-linux-gnuabi64
+        - mips64el-unknown-linux-gnuabi64
+        - powerpc-unknown-linux-gnu
+        - powerpc64-unknown-linux-gnu
+        - powerpc64le-unknown-linux-gnu
+        - x86_64-unknown-linux-gnu
+    steps:
+    - uses: actions/checkout@master
+    - name: Install Rust (
```
Lokathor: was there supposed to be a note here after the `(` ?

Lokathor:
renamed?

alexcrichton:
It looks like there's errors on github's end, so I've opened a ticket to see if they can help diagnose these errors.

PR #229, opened 2019-10-04
Fix sincosf(PI)

### LINKED ISSUE #228: sincosf returns (0, 1) on PI instead of (0, -1)
I'm currently porting euclid to no_std and I noticed that the test for 180 degree rotations seems to fail. The problem does indeed seem to be due to the sincosf implementation in libm which returns (0, 1) on sincosf(PI) instead of (0, -1). I didn't test much further yet. sincos seems unaffected. cosf itself also returns the proper -1.

Update: It calls k_cosf(PI - S2PIO2) where S2PIO2 = PI, therefore it simplifies to k_cosf(0.0), which returns 1, which is then directly returned without modification, which seems wrong.

CryZe:
Looks like the implementation was not ported correctly. Some negations were forgotten in a certain branch. Here is the original code in musl that has the negations:

https://github.com/bpowers/musl/blob/94cb2ec2a0ffcb47d24dbf7a30e462505396cf54/src/math/sincosf.c#L66-L67

Resolves #228

### DIFF
--- a/src/math/sincosf.rs
+++ b/src/math/sincosf.rs
@@ -65,11 +65,11 @@ pub fn sincosf(x: f32) -> (f32, f32) {
         /* -sin(x+c) is not correct if x+c could be 0: -0 vs +0 */
         else {
             if sign {
-                s = k_sinf((x + S2PIO2) as f64);
-                c = k_cosf((x + S2PIO2) as f64);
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks! Can you be sure to add some tests for this as well?

CryZe:
Alright, I added a test.

PR #233, opened 2019-11-22
Fix sincosf for interval (7*pi/4, 9*pi/4)

### LINKED ISSUE #232: sincosf returns wrong results on interval (7*pi/4, 9*pi/4)
Using `libm` 0.2.0 / master `a5e39d8f1ad2ad669e62565f98f1dc86ea5fed96`, this simple test case exercising `2*pi` symmetry:
```
#[cfg(test)]
mod tests {
    use libm::sincosf;

    fn d2r(d: f32) -> f32 {
        use std::f32::consts::PI;
        d * PI / 180.
    }

    #[test]
    fn test_sincosf() {
        assert_eq!(sincosf(d2r(330.)), sincosf(d2r(330. - 360.)));
    }
}
```
Fails (for non-limited-precision reasons), with the message:
```
---- tests::test_sincosf stdout ----
thread 'tests::test_sincosf' panicked at 'assertion failed: `(left == right)`
  left: `(0.86602545, -0.4999999)`,
 right: `(-0.5, 0.8660254)`', src/lib.rs:12:9
```
Precision issues aside, note that the sin/cos components are reversed.

Some quick binary searching shows that this condition persists across the interval of [316, 404] degrees, which is to say it appears to affect (7*pi/4, 9*pi/4) rad.

ajfrantz:
A mistake was made in porting musl's implementation which caused the sin
and cos components to be reversed.

closes #232

### DIFF
--- a/src/math/sincosf.rs
+++ b/src/math/sincosf.rs
@@ -89,11 +89,11 @@ pub fn sincosf(x: f32) -> (f32, f32) {
             }
         } else {
             if sign {
-                s = k_cosf((x + S4PIO2) as f64);
-                c = k_sinf((x + S4PIO2) as f64);
+                s = k_sinf((x + S4PIO2) as f64);
+                c = k_cosf((x + S4PIO2) as f64);
             } else {
-                s = k_cosf((x - S4PIO2) as f64);
-                c = k_sinf((x - S4PIO2) as f64);
+                s = k_sinf((x - S4PIO2) as f64);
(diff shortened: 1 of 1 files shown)


Lokathor:
I'm still new as a maintainer so I'll let Alex double check this before it's merged. Either way they'll need to be the one to publish the `0.2.1`

alexcrichton:
Thanks!

PR #237, opened 2020-01-27
add relibm crate

augustin-cheron:
Add a crate to build a c ABI compatible libm in order to test it against other libm test suite.
This crate currently support libc-test use in musl and openlibm tests and benchmark.


tgross35:
Hey! Sorry this hasn't been looked at in a while. If you are still interested in adding this functionality I think we could do this, but it doesn't need to be a separate crate. Instead, just add a `src/no_mangle.rs` file that wraps each function in a `#[no_mangle]` wrapper (use a macro similar to `libm_helper` rather than enumerating everything manually) and is gated behind a new `no-mangle` Cargo feature. Running libc-test sounds like a good idea, but that can come as a follow up.

Also fyi this PR has a checked in binary.

tgross35:
We merged this crate into the https://github.com/rust-lang/compiler-builtins repo, so please feel free to resubmit there with the above changes.

PR #240, opened 2020-04-08
Port newlib hypotf

### LINKED ISSUE #137: port newlib's hypotf
Source: https://sourceware.org/git/gitweb.cgi?p=newlib-cygwin.git;a=blob;f=newlib/libm/mathfp/ef_hypot.c;hb=HEAD

What needs to be done:

ChiefMilesEdgeworth:
Closes #137 

* Ports the newlib version of hypotf into the main module
* Moves the old code into a submodule `musl`
* Adds tests for hypotf

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -26,7 +26,7 @@ mod musl_reference_tests {
 
     // These files are all internal functions or otherwise miscellaneous, not
     // defining a function we want to test.
-    const IGNORED_FILES: &[&str] = &["fenv.rs"];
+    const IGNORED_FILES: &[&str] = &["fenv.rs", "musl/"];
 
     struct Function {
         name: String,
--- a/src/math/hypotf.rs
+++ b/src/math/hypotf.rs
@@ -2,42 +2,149 @@ use core::f32;
 
 use super::sqrtf;
 
+// pub fn hypotf(mut x: f32, mut y: f32) -> f32 {
+//     let x1p90 = f32::from_bits(0x6c800000); // 0x1p90f === 2 ^ 90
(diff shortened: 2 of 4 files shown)


ChiefMilesEdgeworth:
Actually, this has a test error on the `cargo test --features musl-reference-tests` test. 
This has to do with the new directory added, and I'm not sure what I could do to fix that. I'll dig around, but if someone knows where the issue is I'll accept the help

ChiefMilesEdgeworth:
The test still fails, but now it's a docker issue, so I have no idea.

ChiefMilesEdgeworth:
Here's the error:
```
error[E0463]: can't find crate for `std`
  |
  = note: the `x86_64-unknown-linux-musl` target may not be installed

error: aborting due to previous error

For more information about this error, try `rustc --explain E0463`.
thread 'main' panicked at 'assertion failed: status.success()', build.rs:344:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

alexcrichton:
Thanks! If we switch the implementations then there's no need to keep the old implementation. Is this implementation better than the old one? (e.g. faster, more accurate, etc)

ChiefMilesEdgeworth:
The reason for porting this implementation as I understand it was to avoid using `f64`s. I'm not sure how to benchmark the speed differences, but the theory is that this will run significantly faster on architectures without a builtin representation for `f64` types. I think most of the discussion on this was in the issue this PR closes, #137

tgross35:
Sorry this PR has been around so long. I think that similar to https://github.com/rust-lang/libm/pull/169 there may be some licensing concerns here, but we might have an alternative solution.

This crate has also since been merged into the https://github.com/rust-lang/compiler-builtins repo, so I am going to close this.

Issue #215, opened 2019-08-03
Package is incorrectly licensed Apache/MIT


brson:
libm is derivative of musl, which is MIT licensed, so this source code should also be MIT licensed. Cargo.toml though says "MIT or Apache-2.0". As suggested by the "or", I don't believe it is ok to remove the MIT license and instead apply the Apache license, when the original source is explicitly MIT only.

I am not entirely confident about how the "or" relationship, and adding more restrictions via apache, interacts with the original work though. The obviously correct thing to do is to remove the Apache license and preserve the license of the original work.

workingjubilee:
The Expat ("MIT") License explicitly includes the power to use, copy, modify, merge, publish, distribute, **sublicense**, and/or sell copies of the Software, and to permit persons to whom the Software is furnished to do so, subject to the usual conditions.

That is how people can relicense Expat Licensed code as proprietary.
The original notice must remain, of course.

Unless I'm missing something, Apache 2.0 does not grant any power that the Expat License does not grant and imposes some additional restrictions besides, so it is a subset of the Expat License's permissions, and thus fully covered by the sublicensing claused.

gnzlbg:
@brson 

> ibm is derivative of musl, which is MIT licensed,

Are derivative works required to maintain the original license ? `libm` is a rewrite of `musl` in Rust, that's derivative work, but with substantial changes.

workingjubilee:
I believe the answer would be "No", on reflection, because asserting it must be is like asserting that because, hmm, E.L. James' 50 Shades of Grey is Twilight fanfic with the serial numbers rubbed off, that Stephanie Meyer is entitled to royalties. Notably, no one has tried to pursue this case, in spite of the means and motive being present.

On the other hand, Marvin Gaye's estate vs. Robin Thicke and Pharrell Williams.
So, that is a question best answered by a lawyer.

Lokathor:
Glancing through [this blog post](https://writing.kemitchell.com/2016/09/21/MIT-License-Line-by-Line.html), which was written by _a_ lawyer who is not _our_ lawyer, the MIT license doesn't appear to enforce that the derivative work is also MIT licensed the same way that (for example) GPL does.

infinity0:
Additionally, all the .rs files have, and only have, Sun Microsystems copyright headers with a custom license like this:

~~~~
 * Copyright (C) 1993 by Sun Microsystems, Inc. All rights reserved.
 *
[...]

PR #241, opened 2020-04-13
Remove comments that don't fit our license

Lokathor:
This should resolve https://github.com/rust-lang/libm/issues/215

Per the [latest comment](https://github.com/rust-lang/libm/issues/215#issuecomment-596256963) on the issue, we just have to remove the _comments_, we don't have to actually change our rust code.

I've let stay:
* Comments indicating where the ported code came from, which naturally must have been written by someone doing the rust work.
* Comments that are a math expression without any English (which should be "a fact" and not copyright-able, i think).
* Comments that were obviously added by someone writing the Rust, such as FIXME/TODO, explanations of when we're trying to get LLVM to do the right thing, etc.

Also I've added a blank `rustfml.toml` which should ensure that the crate always uses the default format settings.

This should be ready for review. I may have removed too much or too little of the comments, but there should be no code changes.


BartMassey:
Rather than this, I would prefer to see #241 resolved as [suggested](https://github.com/rust-lang/libm/issues/215#issuecomment-782475374) by @joshtriplett there: put the original copyright notices back.

joshtriplett:
We should not be trying to delete comments in order to get out of preserving copyright notices. We should just preserve the copyright notices, and put them in the COPYRIGHT file.

PR #244, opened 2020-05-29
Use macros for more division/array checks

alexcrichton:
This commit moves over more array accesses to the `i!` macro to avoid
bounds checks when debug assertions are disabled. This is surfaced from
rust-lang/compiler-builtins#360 where recent changes in codegen units
has caused some bounds checks to not get elided in release mode. This
also adds a `div!` macro to work around rust-lang/rust#72751.

### DIFF
--- a/crates/compiler-builtins-smoke-test/Cargo.toml
+++ b/crates/compiler-builtins-smoke-test/Cargo.toml
@@ -3,4 +3,7 @@ name = "cb"
 version = "0.1.0"
 authors = ["Jorge Aparicio <jorge@japaric.io>"]
 
-[dependencies]
+[lib]
+test = false
+bench = false
+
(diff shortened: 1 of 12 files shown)


Lokathor:
The `div!` macro should likely not contain the `unsafe{ ... }` block. That should be left to users of the macro to check for the rhs not being zero on each usage.

Also the div macro should have some sort of doc string about why a person would care to use it, probably with a link to that 72751 issue.

alexcrichton:
That's not really how this is degined to be used. Like the [`i!` macro definition](https://github.com/rust-lang/libm/blob/3d729b7a85daad3b4d54b22e02d00681762bc602/src/math/mod.rs#L9-L37) the unsafety is encapsulated in the macro, rightfully so because it's supposed to be a safe macro. The only purpose is to cater to compiler-builtins which can't panic in release mode, so we're relying on testing elsewhere to weed out possible cases where it does indeed panic or cause unsafety.

Lokathor:
okay, i see.

You should put that explanation as a comment then.

PR #251, opened 2021-03-28
add "readme"-key to Cargo.toml in order for this crate to have a preview on crates.io

phip1611:
Right now there is no nice preview/readme on crates.io. This PR fixes this (with the next release that contains this change.)


Lokathor:
good point. thanks.

while this has failures, they're surely unrelated to this pr because this pr doesn't touch code.

Lokathor (approved):

PR #252, opened 2021-06-22
Fix substract with borrow in FMA

### LINKED ISSUE #242: FMA implementation producing incorrect results
I've run into an issue with the FMA implementation when investigating creating a `no_std` version of my double-double arithmetic crate. It breaks the algorithm for multiplying two `f64` values to produce a double-double, for which I am using algorithm 3 from [Joldes et al. (2017)](https://hal.archives-ouvertes.fr/hal-01351529v3/document): the high word of the result is the standard floating point multiplication, while the low word uses fused-multiply-add.

The output of the algorithm should be two non-overlapping words, I'm testing this with code similar to the following (uses `rand` 0.7):

```rust
#[cfg(test)]
mod tests {
    use rand::Rng;

    #[test]
    fn fma_test() {
        let mut rng = rand::thread_rng();
        let dist = rand::distributions::Uniform::<f64>::new_inclusive(-10000.0, 10000.0);

        for _ in 0..1000000 {
            let a = rng.sample(dist);
            let b = rng.sample(dist);
            let hi = a * b;
            let lo = libm::fma(a, b, -hi); // 1

            if !hi.is_normal() || !lo.is_normal() { continue; }
            assert!(
                libm::ilogb(hi) - libm::ilogb(lo) >= f64::MANTISSA_DIGITS as i32,
                "fma({:?}, {:?}, {:?})", a, b, -hi
            );
        }
    }
}
```

The above test fails with `libm` 0.2.1. If the line marked `// 1` is replaced with

```rust
            let lo = a.mul_add(b, -hi);
```

then the test passes.

The equivalent code for `f32` using `libm::fmaf` passes.

I'm not sure if this is the same as #200 but so far I've not seen segfaults.

jethrogb:
Fixes #242

I assume there's a good reason we're not just using u128 math here.

### DIFF
--- a/src/math/fma.rs
+++ b/src/math/fma.rs
@@ -122,9 +122,9 @@ pub fn fma(x: f64, y: f64, z: f64) -> f64 {
         rhi += zhi + (rlo < zlo) as u64;
     } else {
         /* r -= z */
-        let t = rlo;
-        rlo = rlo.wrapping_sub(zlo);
-        rhi = rhi.wrapping_sub(zhi.wrapping_sub((t < rlo) as u64));
+        let (res, borrow) = rlo.overflowing_sub(zlo);
+        rlo = res;
+        rhi = rhi.wrapping_sub(zhi.wrapping_add(borrow as u64));
         if (rhi >> 63) != 0 {
             rlo = (-(rlo as i64)) as u64;
             rhi = (-(rhi as i64)) as u64 - (rlo != 0) as u64;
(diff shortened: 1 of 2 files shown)


jethrogb:
@Lokathor could you review this? This is resulting in [`std` test failures](https://github.com/rust-lang/rust/pull/85925#discussion_r654904770) on tier 2 targets.

Lokathor:
This change looks reasonable at a glance.

I know the test failures aren't part of this PR, but if you could fix them up that would be appreciated. I can have a proper look at things in about 10hrs from now.

jethrogb:
Pushed some changes

src/math/fma.rs
```diff
@@ -218,6 +218,14 @@ mod tests {
             -0.00000000000000022204460492503126,
         );
 
-        assert_eq!(fma(-0.992, -0.992, -0.992), -0.00793599999988632,);
+        assert_eq!(fma(-0.992, -0.992, -0.992), -0.007936000000000007,);
```
jethrogb, line 221: This is the result from GNU libm

jethrogb:
The following tests also fail on master:
* lgammaf_matches_musl
* tgamma_matches_musl
* y0f_matches_musl
* y0_matches_musl
* lgamma_matches_musl

Fixing these seems like a rabbit hole that I don't want to do.

Lokathor:
Oh boy. Yeah none of that is your fault, I thought it was just some extra warnings and stuff.

Lokathor (approved):

Issue #154, opened 2018-10-17
Float round and roundf are slow


raphlinus:
This bug is branched from rust-lang/rust#55107.

The code for round (adapted from musl libm) is correct (I verified it for all f32 values) but slow, containing many branches.

The following code is also correct and should be much faster (this is the f32 version, which is the main thing I'm testing now, since I'm looking at simd). This is the positive branch.

```rust
    ((x + (0.25 - 0.5 * f32::EPSILON)) + (0.25 + 0.5 * f32::EPSILON)).floor()
```

There are two ways to make this work for signed values: one is to branch on the sign of the argument, the other is to do `copysign(... x.abs() ..., x)`. The copysign intrinsic is not currently exposed in the Rust std lib though, for which I'll file an issue. One of the nice things about copysign is that there's a [wasm intrinsic](https://github.com/sunfishcode/wasm-reference-manual/blob/master/WebAssembly.md#floating-point-copysign). That should get this function down to around 9 asm instructions from the current ~92.

It would also be really _nice_ to expose the [`nearest`](https://github.com/sunfishcode/wasm-reference-manual/blob/master/WebAssembly.md#floating-point-round-to-nearest-integer) wasm instruction as well; the problem is that there's no llvm intrinsic with the exactly correct semantics. I'm going to be exploring that as well.

tgross35:
Looks like this was done in https://github.com/rust-lang/libm/pull/253, this was just never closed.

PR #253, opened 2021-07-01
Optimize round and roundf

bvinc:
In reference to #154, this optimizes the `round` and `roundf` functions.  This does not change the behavior at all.

I compared all f32 values to verify that the behavior of `roundf` has not changed.

For `round`, I directly translated the `roundf` function and verified that it behaved the same with random bits for the sign and exponent and beginning and end of the mantissa.

I ran some criterion benchmarks and basically both functions went from about 7ns to about 3ns.

```
round/round             time:   [7.1414 ns 7.1470 ns 7.1543 ns]                         
round/round_new         time:   [3.0178 ns 3.0311 ns 3.0471 ns]                             

roundf/roundf           time:   [6.8636 ns 6.8767 ns 6.9016 ns]                           
roundf/roundf_new       time:   [2.8326 ns 2.8401 ns 2.8562 ns]                               
```

### DIFF
--- a/src/math/round.rs
+++ b/src/math/round.rs
@@ -1,38 +1,10 @@
+use super::copysign;
+use super::trunc;
 use core::f64;
 
-const TOINT: f64 = 1.0 / f64::EPSILON;
-
 #[cfg_attr(all(test, assert_no_panic), no_panic::no_panic)]
-pub fn round(mut x: f64) -> f64 {
-    let i = x.to_bits();
-    let e: u64 = i >> 52 & 0x7ff;
-    let mut y: f64;
-
-    if e >= 0x3ff + 52 {
-        return x;
-    }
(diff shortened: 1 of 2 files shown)


Lokathor:
Looks like we had some CI trouble, but actually we haven't been passing CI lately to begin with.

It appears that none of the failures were because of thia PR.

My only thought is that in general we might want to mark these and other functions as `inline`, but that can be done later.

@bvinc I can merge if you're all done.

bvinc:
I'm done, but I'll add #inline too if you want.  Looks like it should definitely be on copysign/copysignf.  Are you sure we also want inline on trunc/truncf?

Lokathor:
oh i forgot how big that one is.

we'll do inline separate, I'll just merge for now.

PR #262, opened 2022-03-10
[Blocked] Use const fn where easily applicable.

phip1611:
**Update 2023-12-18**

This is still blocked by having these features not in stable.

```
#![feature(const_float_bits_conv)]
#![feature(const_fn_floating_point_arithmetic)]
#![feature(const_float_classify)]
```

---

This commit adds "const fn" where possible without major refactorings. All functions that rely on special LLVM runtime intrinsic (such as SSE) or `core::ptr::read_volatile` are untouched by this.

Currently, the CI does not pass. I'd like to know if you think this PR is beneficial at all. If so, we/I can work further on it. Currently, many functions can't be ported to be a "const fn" easily because many of them rely on `core::ptr::read_volatile` during runtime.

This will be breaking because it requires a recent rust nightly. Let me know what you think!

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -11,6 +11,10 @@
 #![allow(clippy::float_cmp)]
 #![allow(clippy::eq_op)]
 #![allow(clippy::assign_op_pattern)]
+#![deny(clippy::missing_const_for_fn)]
+#![feature(const_float_bits_conv)]
+#![feature(const_fn_floating_point_arithmetic)]
+#![feature(const_float_classify)]
 
 mod math;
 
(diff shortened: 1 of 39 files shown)


Amanieu:
I am generally in favor of the PR!

The volatile reads are a bit of a hack. I believe they were used for 2 purposes:
- To set the appropriate FP status flags (overflow, underflow, etc) in the FP status register. We really shouldn't bother with this since we don't guarantee any particular value for flags.
- To work around FP precision issues on 32-bit x86 which uses 80-bit float precision internally. This is unfortunately an unavoidable issue on x86 since it has pretty broken FP support.

PR #265, opened 2022-07-30
Ensure more functions do not panic

ankane:
This PR:

- adds `#[cfg_attr(all(test, assert_no_panic), no_panic::no_panic)]` to more functions
- removes panics from `exp10` and `exp10f`

This leaves `frexp` and `frexpf` as the only remaining functions that can panic (wasn't able to figure out how to fix).

Edit: it looks like `j1f` (and possibly `j1` can also panic)

Edit 2: Not sure why, but it looks like the tests didn't catch [integer division](https://github.com/rust-lang/libm/blob/d6d8a1610c9912c84c4c1d12d994d0204dc702ef/src/math/lgamma_r.rs#L155) in `lgamma_r`/`lgammaf_r`

### DIFF
--- a/src/math/acosh.rs
+++ b/src/math/acosh.rs
@@ -7,6 +7,7 @@ const LN2: f64 = 0.693147180559945309417232121458176568; /* 0x3fe62e42,  0xfefa3
 /// Calculates the inverse hyperbolic cosine of `x`.
 /// Is defined as `log(x + sqrt(x*x-1))`.
 /// `x` must be a number greater than or equal to 1.
+#[cfg_attr(all(test, assert_no_panic), no_panic::no_panic)]
 pub fn acosh(x: f64) -> f64 {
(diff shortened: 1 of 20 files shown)


Amanieu:
`j1f` and `j1` seem to have indexing which might be calling panics.

ankane:
Quick update: From what I can tell, the panics for `frexp`/`frexpf` and `j1`/`j1f` are due to recursion. The tests pass for `j1` but not `j1f` since `j1f` has this [additional test](https://github.com/rust-lang/libm/blob/d6d8a1610c9912c84c4c1d12d994d0204dc702ef/src/math/j1f.rs#L365-L369). Other functions with recursion like `j0` and `j0f` also likely can panic, but the tests don't catch it.

Amanieu:
Thanks!

ankane:
Thanks @Amanieu!

PR #272, opened 2022-11-07
Add `rint` and `rintf`

Jules-Bertholet:
Necessary to support https://github.com/rust-lang/rust/pull/95317


Jules-Bertholet:
Actually, I think it might make more sense to just use `rint`/`rintf`...

Amanieu:
The differences between `rint` and `roundeven` are:
- `rint` uses the rounding mode from the FP environment (which defaults to round-to-nearest) while `roundeven` always rounds to nearest even.
- `rint` raises `FE_INEXACT` if rounding occurs, `roundeven` doesn't.

As you said though, Rust already assumes the default rounding mode, so it doesn't matter too much. I think `roundeven` is the "more correct" choice to use but it's not well supported so it's OK to just use `rint` for now. Ideally we should support both in libm.

Jules-Bertholet:
This change needs to make its way to a stable release before https://github.com/rust-lang/rust/pull/95317 can be unblocked.

Amanieu:
I published a new version. However you will need to update compiler-builtins which includes libm as a submodule. You'll need to add a re-export here: https://github.com/rust-lang/compiler-builtins/blob/master/src/math.rs

Issue #255, opened 2021-07-15
Proving IEEE-754 2008 compliance for all sqrt/sqrtf implementations on all platforms.


james7132:
Rust's `f32`/`f64` types specify that they're IEEE-754 2008 compliant, from which the behavior of `sqrt` is explicitly defined by the standard. The SSE/SSE2 path in both is currently the only `target-feature` based platform specific alterations in the libm codebase. While these intrinsic based paths are significantly faster, there should be extensive testing to show that these alternative implementations produce identical results across platforms that is consistent with the standard.

For deterministic game networking and simulation stability, we're relying on `libm` to provide bit-for-bit identical output across multiple targets for common mathematical functions that are not consistent across hardware implementations (i.e. much of the x87 transcendental instructions), and `sqrt(f)` is a particularly common function to see used.

Lokathor:
**For Background:** The `f32::sqrt` and `f64::sqrt` methods in `std` use the compiler intrinsic for sqrt. This means that all `sqrt` method calls should naturally utilize hardware as much as possible. This means not only on targets with sse/sse2, but also arm, wasm, etc. The conditional compilation in `libm::sqrtf` and `libm::sqrt` does use a CPU intrinsic when statically available. However, that code path only exists as an optimization to be used by `no_std` libraries that need to do a sqrt and so are calling `libm` directly. Uses of the `sqrt` methods in `std` should never end up lowering into a `libm` call if the hardware supports a sqrt operation.

That said, we welcome improvements to the crate's test suite.

tgross35:
`sqrtf` is now tested exhaustively against MPFR and passes 🎉 , `sqrt` has extensive tests against the same.

PR #274, opened 2022-11-14
Add a small conformance test to sqrt/sqrtf

ZagButNoZig:
**What this PR does**

Add a test that guarantees that the result of sqrt/sqrtf is bit for bit identical on all platforms for a given array of test values.

**Why this PR**

This is a proposed fix for issue #255. Which asks all sqrt implementations to be IEEE 754 2008 compliant. While this does not prove compliance with the spec it at least proves conformance between platforms.

**Issues with this PR**

Since the values and their results are hard coded, they would need to be updated on every implementation change. The values are also somewhat arbitrary and don't cover every case inside the sqrt implementation.

Any feedback is very welcome.

### DIFF
--- a/src/math/sqrt.rs
+++ b/src/math/sqrt.rs
@@ -261,16 +261,20 @@ mod tests {
             assert_eq!(sqrt(f), f);
         }
     }
-    
+
     #[test]
     fn conformance_tests() {
         let values = [3.14159265359, 10000.0, f64::from_bits(0x0000000f), INFINITY];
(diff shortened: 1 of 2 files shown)


Amanieu:
Tests are failing for sqrt(-1) because the sign of the NaN is different from what was expected. I think this is an acceptable divergence.

Amanieu:
The sign of a NaN result is implementation-defined and not specified by IEEE-754. So it makes sense that different architectures produce different results.

PR #275, opened 2023-01-10
Add a generic helper over f32/f64

moritz-meier:
This PR adds a generic `Libm<T>` helper to abstract over f32 and f64.
With this you can call the libm functions like this `libm::Libm::<f32>::acos(x)` and `libm::Libm::<f64>::acos(x)`.
I think it should all be zero-cost abstraction.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,21 @@
+# EditorConfig helps developers define and maintain consistent
+# coding styles between different editors and IDEs
+# editorconfig.org
+
+root = true
+
+[*]
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+indent_style = space
+indent_size = 4
+
+[*.md]
+# double whitespace at end of line
+# denotes a line break in Markdown
+trim_trailing_whitespace = false
+
+[*.yml]
+indent_size = 2
--- a/src/libm_helper.rs
+++ b/src/libm_helper.rs
@@ -1,0 +1,171 @@
+use core::marker::PhantomData;
+
+use crate::*;
+
+/// Generic helper for libm functions, abstracting over f32 and f64. <br/>
+/// # Type Parameter:
+/// - `T`: Either `f32` or `f64`
+///
+/// # Examples
+/// ```rust
+/// use libm::{self, Libm};
+///
+/// const PI_F32: f32 = 3.1415927410e+00;
+/// const PI_F64: f64 = 3.1415926535897931160e+00;
+///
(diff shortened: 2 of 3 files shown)


Amanieu:
Are you using this on an embedded target? Otherwise I see no reason why you wouldn't just use the standard library methods on `f32` and `f64` directly.

moritz-meier:
yes, libm is used in the uom crate for no_std builds. In the uom tests quickcheck is used which is generic over f32/f64 thus we need also generic libm methods.

https://github.com/iliekturtles/uom/pull/309

src/libm_helper.rs
```diff
@@ -0,0 +1,143 @@
+use core::marker::PhantomData;
+
+use crate::*;
+
+pub struct Libm<T>(PhantomData<T>);
```
Amanieu, line 5: Add some doc comments here to explain what this type is and some examples of how it can be used.

src/libm_helper.rs
```diff
@@ -0,0 +1,143 @@
...
+			#![allow(unused_parens)]
+
+			$(
+				#[inline(always)]
+				pub fn $func($($arg: $arg_type),*) -> ($($ret_type),*) {
+					$libm_fn($($arg),*)
+				}
+			)*
+		}
+	};
```
Amanieu, line 20: Can you fix the formatting here? Use spaces instead of tabs, and remove the blank lines.

src/libm_helper.rs
```diff
@@ -0,0 +1,143 @@
+use core::marker::PhantomData;
+
+use crate::*;
+
+pub struct Libm<T>(PhantomData<T>);
+
+macro_rules! libm_helper {
+	($t:ident, $(($libm_fn:ident => $func:ident($($arg:ident: $arg_type:ty),*) -> ($($ret_type:ty),*)));*) => {
```
Amanieu, line 8: Can you spread this out over multiple lines to make it easier to read?

.editorconfig
```diff
@@ -0,0 +1,21 @@
...
+indent_style = space
+indent_size = 4
+
+[*.md]
+# double whitespace at end of line
+# denotes a line break in Markdown
+trim_trailing_whitespace = false
+
+[*.yml]
+indent_size = 2
```
Amanieu, line 21: Don't commit your editorconfig.

Amanieu (approved):

.editorconfig
moritz-meier: Because you mentioned it earlier, it would be nice to configure and document the suggested format in the project.
I copied the .editorconfig from the rust project.

PR #282, opened 2023-08-02
Exclude ci directory from packaged crate

mulkieran:


mulkieran:
@Amanieu Would you consider this change? I've explained my reasoning in the commit message and I think it is good.

Amanieu:
I don't normally look at commit messages, only the PR summary. But the arguments are reasonable.

Could you also exclude `.github/workflows`? That shouldn't be included either.

mulkieran:
> I don't normally look at commit messages, only the PR summary. But the arguments are reasonable.
> 
> Could you also exclude `.github/workflows`? That shouldn't be included either.

Done!

PR #288, opened 2023-12-18
remove force_eval! macro, prepare more const functions, fix clippy, and add clippy to CI

phip1611:
This MR is intended to be a preparation for #262.

- fix all clippy warnings (or properly ignore, where sensible)
- add clippy to CI
- remove the `force_eval!` macro with something more `const` friendly

Regarding the removal of the `force_eval!`-macro: 
Fomment from @Amanieu from the other PR:

> I am generally in favor of the PR!
> 
> The volatile reads are a bit of a hack. I believe they were used for 2 purposes:
> 
> * To set the appropriate FP status flags (overflow, underflow, etc) in the FP status register. We really shouldn't bother with this since we don't guarantee any particular value for flags.
> * To work around FP precision issues on 32-bit x86 which uses 80-bit float precision internally. This is unfortunately an unavoidable issue on x86 since it has pretty broken FP support.


phip1611:
ping @Amanieu

Amanieu:
The test failures don't happen on master, this is definitely related to your changes.

Amanieu:
You can reproduce the test failures locally with `cargo test --release`.

phip1611:
I'd love to have this, but I don't have the

PR #289, opened 2024-02-11
Remove unneeded `extern core` in `tgamma`

Amjad50:
This was causing some issues when building the crate for different targets that don't have `core` builtin.


Amjad50:
Closed because of #290

Amanieu:
I think this can be merged if you re-open it. You can also fix the CI issues by just allowing the internal features warning.

Amjad50:
Alright, I reopened this now. I'll probably close the other PR if weak symbols are a better approach

Amjad50:
Thanks @Amanieu

PR #290, opened 2024-02-18
Compiling `libm` for `std`

Amjad50:
Added the ability to compile `libm` as a child dependency of `std`.

This will allow fixing the issue where some targets that don't have `libc` linked can have math functions without linking errors.

This fixes the issue originating in https://github.com/rust-lang/compiler-builtins/issues/576

This also replaces #289 as its included in this one.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,6 +12,12 @@ version = "0.2.8"
 edition = "2018"
 exclude = ["/ci/", "/.github/workflows/"]
 
+[dependencies]
+# Internal feature, only used when building as part of libstd, not part of the
+# stable interface of this crate.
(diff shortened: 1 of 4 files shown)


Amjad50:
The compilation fails because of usage of internal feature

Amanieu:
I think a better approach would be to have compiler-builtins *always* create the necessary symbols on all targets, but define them as weak symbols. There is already functionality for this in compiler-builtins, we just need to make sure it is disabled for targets that don't properly support weak symbols (Apple & Windows).

The reasoning behind this is to allow using platform-provided optimized versions of the math functions when they are available since they can be faster than our implementation in libm.

Amjad50:
Ah I didn't know about weak symbols, I'll look into it and add it there then.

Thanks

PR #292, opened 2024-04-02
Start removing force_eval uses

saethlin:
I'm starting here by deleting all the `force_eval` uses that are for raising floating-point exceptions.

@programmerjake is interested in some breadcrumbs for the non-default rounding modes uses, so I'm going to do something different for those.


saethlin:
`force_eval!(0.0 / 0.0)` appears to not do anything at all:
https://play.rust-lang.org/?version=nightly&mode=release&edition=2021&gist=2621c439604f11b4ef53462f228f8c33
```rust
fn main() {
    unsafe {
        println!("{:032b}", _mm_getcsr());
        force_eval!(0.0 / 0.0);
        println!("{:032b}", _mm_getcsr());
    }
}
```
prints the same value twice.

saethlin:
Since there's now effort port new implementations from musl, I'm more against changes like this that would cause implementation divergence.

PR #294, opened 2024-04-11
`TotalEq` -0 in `fabs` test

Rudxain:
`assert_eq!` uses `PartialEq`, which doesn't discern between -0 and +0. `total_cmp` does distinguish them. So this patch makes the test spec-compliant

### DIFF
--- a/src/math/fabs.rs
+++ b/src/math/fabs.rs
@@ -32,7 +32,7 @@ mod tests {
     fn spec_tests() {
         assert!(fabs(NAN).is_nan());
         for f in [0.0, -0.0].iter().copied() {
-            assert_eq!(fabs(f), 0.0);
+            assert_eq!(fabs(f).total_cmp(&0.0), core::cmp::Ordering::Equal);
         }
         for f in [INFINITY, NEG_INFINITY].iter().copied() {
             assert_eq!(fabs(f), INFINITY);


Rudxain:
I recommend reviewing any other code that uses `==` and/or `assert_eq!`, to check if it should be replaced by [`total_cmp`](https://doc.rust-lang.org/core/primitive.f64.html#method.total_cmp). `total_cmp` also distinguishes between signed `NAN`s, which may be useful somewhere else in the codebase

Amanieu:
It would be better to use `.to_bits()` to ensure an exact match.

Rudxain:
> `.to_bits()`

I agree, but that would only apply to other cases (such as performance-sensitive code), not here. The spec imposes no requirements about the bit pattern

Amanieu:
There is only a single bit pattern for 0 or -0.

Rudxain:
> There is only a single bit pattern for 0 or -0.

Yes, but some platforms use "flush-to-zero" for subnormal numbers. I'm not entirely sure how that mechanism works 🤔

Amanieu:
That doesn't have any effect here. It's only done during arithmetic operations, not comparisons. In any case the Rust language doesn't allow flush-to-zero behavior, so such modes must be turned off in the CPU when running Rust code.

tgross35:
Tests for fabs are now using assert_biteq which checks the exact repr, so this shouldn't be needed anymore https://github.com/rust-lang/libm/blob/932898e0c18ad63bf209d8ddda2417a5196f15da/src/math/fabs.rs#L58-L61.

Thank you for the PR!

PR #297, opened 2024-08-06
Add `force-soft-floats` feature to prevent using any intrinsics or arch-specific code

GuillaumeGomez:
It comes from this [zulip discussion](https://rust-lang.zulipchat.com/#narrow/stream/182449-t-compiler.2Fhelp/topic/Infinite.20recursion.20in.20compiler-builtins) which is blocking the last GCC backend sync with rustc.

cc @Amanieu

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,9 @@ unstable = []
 # musl libc.
 musl-reference-tests = ['rand']
 
+# Used to prevent using any intrinsics or arch-specific code.
+only-soft-floats = []
+
 [workspace]
 members = [
   "crates/compiler-builtins-smoke-test",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,7 +24,7 @@ unstable = []
 musl-reference-tests = ['rand']
 
 # Used to prevent using any intrinsics or arch-specific code.
-only-soft-floats = []
+force-soft-floats = []
 
 [workspace]
 members = [
--- a/crates/compiler-builtins-smoke-test/Cargo.toml
+++ b/crates/compiler-builtins-smoke-test/Cargo.toml
@@ -10,4 +10,4 @@ bench = false
 [features]
 unstable = []
 checked = []
-only-soft-floats = []
+force-soft-floats = []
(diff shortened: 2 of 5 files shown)


src/math/ceil.rs
```diff
@@ -12,7 +12,7 @@ pub fn ceil(x: f64) -> f64 {
     // `f64.ceil` native instruction, so we can leverage this for both code size
     // and speed.
     llvm_intrinsically_optimized! {
-        #[cfg(target_arch = "wasm32")] {
+        #[cfg(all(target_arch = "wasm32", not(feature = "only-soft-floats")))] {
```
thomcc, line 15: Should this be done in the `llvm_intrinsically_optimized!` macro definition instead?

GuillaumeGomez: Seems like this `cfg` (`target_arch = wasm32`) is always used in this macro call, so might as well move it directly into it, good idea!

thomcc: Oh, I meant the `not(feature = "only-soft-floats")` bit, but I don't have strong feelings either way.

GuillaumeGomez:
Applied suggestion!

src/math/ceil.rs
GuillaumeGomez: That's actually what I meant too. ^^'

GuillaumeGomez:
And even fixed CI. :)

Cargo.toml
```diff
@@ -23,6 +23,9 @@ unstable = []
 # musl libc.
 musl-reference-tests = ['rand']
 
+# Used to prevent using any intrinsics or arch-specific code.
+only-soft-floats = []
```
Amanieu, line 27: Can you rename this to `force-soft-float`.

GuillaumeGomez: Done!

Amanieu:
Thanks! The next step is to update the libm submodule in compiler-builtins and ensure the feature is enabled.

GuillaumeGomez:
Gonna send a PR in a few hours.

PR #300, opened 2024-10-05
Test more targets against a custom-built musl libm

tgross35:
We don't currently test on any non-linux targets because results always require running against a `linux-musl` target. This PR does the following:

- Add `musl-math-sys` which builds musl's math symbols with `cc` for a better test basis
- Add macros and traits for a generic test framework
- Add a simple test with 5000 random inputs that compares our `libm` against the build musl version
- Now that testing isn't limited to linux, enable CI for all targets tested in `compiler-builtins`

CI testing is still somewhat limited (MSVC and Wasm can't build musl sources) but these changes should be enough to make future improvements easier.

### DIFF
--- a/crates/libm-macros/src/lib.rs
+++ b/crates/libm-macros/src/lib.rs
@@ -1,0 +1,541 @@
+mod parse;
+use std::sync::LazyLock;
+
+use parse::{Invocation, StructuredInput};
+use proc_macro as pm;
+use proc_macro2::{self as pm2, Span};
+use quote::{ToTokens, quote};
+use syn::Ident;
+use syn::visit_mut::VisitMut;
+
+const ALL_FUNCTIONS: &[(Signature, Option<Signature>, &[&str])] = &[
+    (
+        // `fn(f32) -> f32`
+        Signature { args: &[Ty::F32], returns: &[Ty::F32] },
+        None,
+        &[
+            "acosf", "acoshf", "asinf", "asinhf", "atanf", "atanhf", "cbrtf", "ceilf", "cosf",
+            "coshf", "erff", "exp10f", "exp2f", "expf", "expm1f", "fabsf", "floorf", "j0f", "j1f",
+            "lgammaf", "log10f", "log1pf", "log2f", "logf", "rintf", "roundf", "sinf", "sinhf",
+            "sqrtf", "tanf", "tanhf", "tgammaf", "truncf",
+        ],
+    ),
+    (
+        // `(f64) -> f64`
+        Signature { args: &[Ty::F64], returns: &[Ty::F64] },
+        None,
+        &[
+            "acos", "acosh", "asin", "asinh", "atan", "atanh", "cbrt", "ceil", "cos", "cosh",
+            "erf", "exp10", "exp2", "exp", "expm1", "fabs", "floor", "j0", "j1", "lgamma", "log10",
+            "log1p", "log2", "log", "rint", "round", "sin", "sinh", "sqrt", "tan", "tanh",
+            "tgamma", "trunc",
+        ],
+    ),
+    (
+        // `(f32, f32) -> f32`
+        Signature { args: &[Ty::F32, Ty::F32], returns: &[Ty::F32] },
+        None,
+        &[
+            "atan2f",
+            "copysignf",
+            "fdimf",
+            "fmaxf",
+            "fminf",
+            "fmodf",
+            "hypotf",
+            "nextafterf",
+            "powf",
+            "remainderf",
+        ],
+    ),
+    (
+        // `(f64, f64) -> f64`
+        Signature { args: &[Ty::F64, Ty::F64], returns: &[Ty::F64] },
+        None,
+        &[
+            "atan2",
+            "copysign",
+            "fdim",
+            "fmax",
+            "fmin",
+            "fmod",
+            "hypot",
+            "nextafter",
+            "pow",
+            "remainder",
+        ],
+    ),
+    (
+        // `(f32, f32, f32) -> f32`
+        Signature { args: &[Ty::F32, Ty::F32, Ty::F32], returns: &[Ty::F32] },
+        None,
+        &["fmaf"],
+    ),
+    (
+        // `(f64, f64, f64) -> f64`
+        Signature { args: &[Ty::F64, Ty::F64, Ty::F64], returns: &[Ty::F64] },
+        None,
+        &["fma"],
+    ),
+    (
+        // `(f32) -> i32`
+        Signature { args: &[Ty::F32], returns: &[Ty::I32] },
+        None,
+        &["ilogbf"],
+    ),
+    (
+        // `(f64) -> i32`
+        Signature { args: &[Ty::F64], returns: &[Ty::I32] },
+        None,
+        &["ilogb"],
+    ),
+    (
+        // `(i32, f32) -> f32`
+        Signature { args: &[Ty::I32, Ty::F32], returns: &[Ty::F32] },
+        None,
+        &["jnf"],
+    ),
+    (
+        // `(i32, f64) -> f64`
+        Signature { args: &[Ty::I32, Ty::F64], returns: &[Ty::F64] },
+        None,
+        &["jn"],
+    ),
+    (
+        // `(f32, i32) -> f32`
+        Signature { args: &[Ty::F32, Ty::I32], returns: &[Ty::F32] },
+        None,
+        &["scalbnf", "ldexpf"],
+    ),
+    (
+        // `(f64, i64) -> f64`
+        Signature { args: &[Ty::F64, Ty::I32], returns: &[Ty::F64] },
+        None,
+        &["scalbn", "ldexp"],
+    ),
+    (
+        // `(f32, &mut f32) -> f32` as `(f32) -> (f32, f32)`
+        Signature { args: &[Ty::F32], returns: &[Ty::F32, Ty::F32] },
+        Some(Signature { args: &[Ty::F32, Ty::MutF32], returns: &[Ty::F32] }),
+        &["modff"],
+    ),
+    (
+        // `(f64, &mut f64) -> f64` as  `(f64) -> (f64, f64)`
+        Signature { args: &[Ty::F64], returns: &[Ty::F64, Ty::F64] },
+        Some(Signature { args: &[Ty::F64, Ty::MutF64], returns: &[Ty::F64] }),
+        &["modf"],
+    ),
+    (
+        // `(f32, &mut c_int) -> f32` as `(f32) -> (f32, i32)`
+        Signature { args: &[Ty::F32], returns: &[Ty::F32, Ty::I32] },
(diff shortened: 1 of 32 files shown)


tgross35:
@Amanieu when you get the chance this should be ready for a look. The goal here is mostly to get some easier testing options off the ground, which can be followed up with less random inputs (can probably leverage periodicity/domain for better coverage) and more accurate output checks (checking ULP against MPFR).

The PR is pretty heavy because it adds a lot of framework, to make things easier it is by-commit reviewable.

tgross35:
The rules macro was getting pretty cumbersome to work with so I wound up replacing it with a proc macro.

crates/libm-test/src/lib.rs
```diff
@@ -1 +1,42 @@
...
+}
+
+/// If only a few checks are incorrect, xfail them here rather than skipping the entire test.
+pub fn xfail<F: Float>(actual: F, expected: F, ctx: &CheckCtx) -> bool {
+    match (&ctx.basis, ctx.fname) {
+        // FIXME(correctness): for large negative inputs (e.g.
+        // -1.7976931348623157e308), we return -NaN but musl says +NaN
+        (CheckBasis::Musl, "tgamma" | "tgammaf") if actual.is_nan() && expected.is_nan() => true,
+        _ => false,
+    }
```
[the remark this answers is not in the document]
tgross35: I adjusted this to only check for copysign and abs.

crates/libm-test/src/gen/random.rs
```diff
@@ -0,0 +1,95 @@
...
+        (f64::NEG_INFINITY, f64::NEG_INFINITY, f64::NEG_INFINITY),
+        (f64::MAX, f64::MAX, f64::MAX),
+        (f64::MIN, f64::MIN, f64::MIN),
+        (f64::MIN_POSITIVE, f64::MIN_POSITIVE, f64::MIN_POSITIVE),
+        (f64::NAN, f64::NAN, f64::NAN),
+    ];
+
+    inputs_i32.extend((0..(ntests - inputs_i32.len())).map(|_| rng.gen::<(i32, i32, i32)>()));
+    inputs_f32.extend((0..(ntests - inputs_f32.len())).map(|_| rng.gen::<(f32, f32, f32)>()));
+    inputs_f64.extend((0..(ntests - inputs_f64.len())).map(|_| rng.gen::<(f64, f64, f64)>()));
```
quaternic, line 82: These are generating only floats in the range `0.0..1.0`, and IIRC not even all of those, for efficiency reasons. Generating integers and mapping with `from_bits` is the quick fix, though it's probably not the ideal distribution either. What would actually make sense depends on function quite a bit. As in, you want to give `sqrt` some negative inputs to test that it returns NaN, but benchmarking it with 50% negative inputs doesn't make much sense.

tgross35: Well that is a big oversight on random number generation, thanks for the catch.

This is just meant to be a generator that is quick and easy to review to get the infrastructure merged, even though it isn't a very good test. I am working on something that will be a bit smarter and take function domain, periodicity, inflection points, etc into account, but that will be a follow up https://github.com/tgross35/rust-libm/blob/e8ca082c19972c680f23fed2915d5ca19543f20f/crates/libm-test/src/domain.rs (that file is just the very basics).

It is still meant to work right though, I'll make the randomness change.

quaternic: One idea I've often used to try to generate a better ratio of interesting inputs is to pick the most important bits to vary exhaustively, and fill in the rest from just a few choices. Like for an integer you might generate all combinations of the 4 high bits and 4 low bits, and fill the middle bits from {all 0s, all 1s, a handful of random patterns}. For a float you also want the exponent low bits and mantissa high bits, so it get a bit messy.

tgross35: My complete idea is to combine the following:

* For each "interesting" point, check every value within ~500 ULP. This will catch behavior at zero, infinities, poles, inflection points, and domain boundaries.
* Do something like a logspace within the function's domain or period to get an even distribution for a certain number of iterations.
* For periodic functions, do some kind of check that they repeat (maybe? or check a couple periods toward the infinities).
* Try a number of random outputs that are not in the function's domain.

I'll also add exhaustive tests that check every value within the domain, behind a gate. Plus full exhaustive tests just to have them.

@beetrees was suggesting we could run exhaustive tests in CI if the relevant source file was changed.

tgross35: For reference, I am kind of basing the idea off of what I did for test-float-parse where there is a common interface, then you have a variety of different generators that can be plugged in https://github.com/rust-lang/rust/tree/master/src/etc/test-float-parse/src/gen

quaternic: I wrote a bit of code for the above idea, and with another look it ended up neater than I expected, so maybe this can come in handy:
```rust
/// Iterator of all bitpatterns that are a subset of the mask.
fn bit_subsets(mask: u32) -> impl Iterator<Item = u32> {
    let ignore = !mask;
    assert!(ignore != 0, "mask can't have all bits set");
    let mut counter = ignore.wrapping_sub(1);
    std::iter::from_fn(move || {
        // Incrementing carries through any ignored bits, which
        // are then toggled back on
        counter = counter.checked_add(1)? | ignore;
        Some(counter & !ignore)
    })
}
```

tgross35: Well, fixing the random number generation produced quite a few more failures. It will be more interesting for the MPFR PR, seems like musl has some unusual edge cases here.

crates/libm-test/src/lib.rs
```diff
@@ -1 +1,64 @@
...
+        "asinh" | "asinhf" => 6,
+        "lgamma" | "lgamma_r" | "lgammaf" | "lgammaf_r" => 400,
+        "tanh" | "tanhf" => 4,
+        "tgamma" => 20,
+        "j0" | "j0f" | "j1" | "j1f" => {
+            // Results seem very target-dependent
+            if cfg!(target_arch = "x86_64") { 4000 } else { 800_000 }
+        }
+        "jn" | "jnf" => 1000,
+        "sincosf" => 2000,
```
quaternic, line 36: The problem with `sincosf` actually seems to be a bug in libm's port of musl. I should have a patch for that soonish

tgross35: Thanks! (this is https://github.com/rust-lang/libm/pull/331)

crates/libm-macros/src/lib.rs
```diff
@@ -0,0 +1,524 @@
...
+            format!("only one of `skip` or `only` may be specified"),
+        );
+        return Err(e);
+    }
+
+    if let Some(map) = &input.fn_extra {
+        // If no default is provided, make sure
+        if !map.keys().any(|key| key == "_") {
+            let mut fns_not_covered = Vec::new();
+            for name in ALL_FUNCTIONS_FLAT.iter().map(|func| func.name) {
```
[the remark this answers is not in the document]
tgross35: Good idea, I moved the the collection to this function and check against that now.

crates/libm-macros/src/lib.rs
```diff
@@ -0,0 +1,524 @@
...
+                if map.keys().any(|key| key == name) {
```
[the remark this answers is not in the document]
tgross35: Great eye. This could probably use a test case (as long with the other parts that reject code), but I don't think it is worth setting up trybuild for initial support.

crates/libm-macros/src/lib.rs
```diff
@@ -0,0 +1,524 @@
...
+    if !input.skip.is_empty() && input.only.is_some() {
+        let e = syn::Error::new(
+            input.only_span.unwrap(),
+            format!("only one of `skip` or `only` may be specified"),
+        );
+        return Err(e);
+    }
+
+    if let Some(map) = &input.fn_extra {
+        // If no default is provided, make sure
```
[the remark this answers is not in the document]
tgross35: This is gone with the other refactoring in this area.

crates/libm-test/src/test_traits.rs
```diff
@@ -0,0 +1,251 @@
...
+                return res;
+            }
+
+            // Check when both are NaNs
+            if self.is_nan() && expected.is_nan() {
+                // By default, NaNs have nothing special to check.
+                return Ok(());
+            } else if self.is_nan() || expected.is_nan() {
+                // Check when only one is a NaN
+                bail!("mismatched NaNs")
```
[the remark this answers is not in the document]
tgross35: I just changed this to "real value != NaN", the context should make it clear which side is which.

crates/libm-test/src/special_case.rs
```diff
@@ -0,0 +1,231 @@
...
+            }
+        }
+
+        maybe_check_nan_bits(actual, expected, ctx)
+    }
+}
+
+/// Check NaN bits if the function requires it
+fn maybe_check_nan_bits<F: Float>(actual: F, expected: F, ctx: &CheckCtx) -> Option<TestResult> {
+    if !(ctx.canonical_name == "abs" || ctx.canonical_name == "copysigh") {
```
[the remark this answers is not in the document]
tgross35: Another great catch. I plan to change everything to enum matching at some point to avoid these kind of issues, but that will be a follow up since this PR already has a lot going on.

crates/libm-test/src/special_case.rs
```diff
@@ -0,0 +1,235 @@
...
+    }
+}
+
+/// Check NaN bits if the function requires it
+fn maybe_check_nan_bits<F: Float>(actual: F, expected: F, ctx: &CheckCtx) -> Option<TestResult> {
+    if !(ctx.canonical_name == "fabs" || ctx.canonical_name == "copysign") {
+        return None;
+    }
+
+    // The musl implementations seem to set the top bit of the mantissa for any NaN on i686.
```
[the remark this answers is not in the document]
tgross35: Thanks for the context, updated the comment

crates/libm-test/src/special_case.rs
```diff
@@ -0,0 +1,236 @@
...
+}
+
+/// Check NaN bits if the function requires it
+fn maybe_check_nan_bits<F: Float>(actual: F, expected: F, ctx: &CheckCtx) -> Option<TestResult> {
+    if !(ctx.canonical_name == "fabs" || ctx.canonical_name == "copysign") {
+        return None;
+    }
+
+    // The musl implementations uses x87 instructions on i686 which quiets signaling NaNs
+    if cfg!(target_arch = "x86") && ctx.basis == CheckBasis::Musl && ctx.canonical_name == "fabs" {
```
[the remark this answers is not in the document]
tgross35: Ah right, I wonder if we should call these via an `extern "C"` wrapper.

tgross35:
> Other than that, LGTM 👍

Thanks for the review!

This has been through quite a few cycles, let's get it merged.

tgross35:
Just to be foiled by a Docker 502...

PR #315, opened 2024-10-22
Add a test generator that checks edge cases

tgross35:
This uses the same edge cases as `compiler-builtins`.


tgross35:
This depends on https://github.com/rust-lang/libm/pull/300

PR #316, opened 2024-10-24
Move arch-specific behavior and intrinsics to a separate module

tgross35:
Introduce `math/arch/*` as a home for architecture-specific behavior and intrinsic usage, and move relevant existing code there.

This allows us to declutter things a bit and also commonize the logic of `cfg` flags so we don't accidentally disobey `force-soft-floats`.


tgross35:
The commit history is a mess so please don't merge directly, but this is ready for a look. The diff makes a lot more sense with "hide whitespace" checked.

PR #317, opened 2024-10-25
Update licensing to MIT AND (MIT OR Apache-2.0)

tgross35:
Currently both Cargo.toml and the license files indicate that this library may be used under either MIT or Apache-2.0. However, this is not accurate; since portions of this library were derived from musl libc, which is available under the MIT license, this terms of use for this library must also include use under the MIT license. That is, it is not correct that this library may be used under only the Apache-2.0 license.

Update the SPDX license identifier to `MIT OR (MIT AND Apache-2.0)` to indicate that use must include the MIT license, but to clarify that contributions are made under `MIT OR Apache-2.0`. This is compatible with the current state of this repository since it has always contained both license files, and the `Cargo.toml` license field has indicated `MIT OR Apache-2.0` since it was added.

In accordance with the above, replace the two license files with a combined LICENSE.txt that makes these terms clear and gives attribution to works from which this library is derived.

Fixes: https://github.com/rust-lang/libm/issues/215
Link: https://rust-lang.zulipchat.com/#narrow/channel/335408-foundation/topic/Request.20for.20legal.20team.20input.20on.20crate.20licensing
Link: https://git.musl-libc.org/cgit/musl/tree/COPYRIGHT


tgross35:
Requesting review from @joshtriplett since you have been knowledgeable about all this.

I changed the copyrights to "Rust Developers" @japaric would you be able to ack this since your name is currently on the copyrights?

LICENSE.txt
```diff
@@ -0,0 +1,236 @@
...
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+------------------------------------------------------------------------------
+
+Portions of this software are derived from third-party works licensed under
+terms compatible with the above MIT license:
+
+* musl libc, copyright Rich Felker, et al. https://www.musl-libc.org/ 
+* The CORE-MATH project https://core-math.gitlabpages.inria.fr/
```
tgross35, line 236: We don't yet have anything from CORE-MATH in the repository, but I included it here since I am already starting on the ports.

joshtriplett:
@tgross35 Strictly speaking, you have to preserve *all* copyright notices. For instance, since we're copying code from projects under the MIT license, if the version of the license they're using starts with copyright notices, we need to preserve *all* of those.

For that exact reason, I would recommend that we not add any *more*, "Rust Developers" or otherwise. But we have to preserve all the ones that already exist within the projects we're copying from.

joshtriplett:
Other than the issue of preserving copyright notices (and not adding any more), LGTM.

tgross35:
I did include `Copyright © 2005-2020 Rich Felker, et al.` at the bottom, but made this text more explicit. This is what is included at the top of the MIT license https://git.musl-libc.org/cgit/musl/tree/COPYRIGHT - is this sufficient?

Later in the file it states:

```
Much of the math library code (src/math/* and src/complex/*) is
Copyright © 1993,2004 Sun Microsystems or
Copyright © 2003-2011 David Schultz or
Copyright © 2003-2009 Steven G. Kargl or
Copyright © 2003-2009 Bruce D. Evans or
Copyright © 2008 Stephen L. Moshier or
Copyright © 2017-2018 Arm Limited
and labelled as such in comments in the individual source files. All
have been licensed under extremely permissive terms.
```

But I assume we are probably okay to not include this since we retain the copyright headers in our math source files.

joshtriplett:
The requirement is to preserve all copyright notices you receive. We should not be pruning any. We can preserve them either in the source files or the license, as long as they're preserved. It'd probably be *convenient* for downstream users if they were all in the license, but the requirement on *our* end is just to preserve them all *somewhere*.

Amanieu (approved):

joshtriplett (approved):
This looks correct to me, and it's a massive improvement.

I personally think we should also copy the copyright notices present in source files into the license file, so that they're all in one place, but that doesn't need to be a blocker for *this* commit.

joshtriplett (approved):

tgross35:
> @japaric would you be able to ack this since your name is currently on the copyrights?

As suggested by @joshtriplett, I updated the file to persist this copyright since there are quite a few we have to retain anyway: https://github.com/rust-lang/libm/blob/b8d8354982b4ce0991952ac33bf78b991ad05b4d/LICENSE.txt#L228-L230

Issue #310, opened 2024-10-18
Make functions in this crate `const`?


JSorngard:
Now that `const_float_arithmetic` has been stabilized, will the functions in this crate be made `const` where possible? Or is this not in scope due to MSRV policy?

tgross35:
The stabilization definitely would make it possible for some of these functions to work in `const` contexts. However, I don't think that is an API guarantee we want to make because it would be a pretty significant limit on flexibility (anything that uses SIMD, assembly, or generic algorithms would no longer usable).

I think the better solution is to just move the math functions from `std` to `core`, and then make those `const` using the host machine's implementation. This is already planned but we do need to improve some things in this crate before it is possible (e.g. fix rounding errors, make sure our implementation is performant enough to ship).

tgross35:
I am going to close this since I don't think we will make any changes in this crate's public API. I brought the topic of const float math up on Zulip since I don't think it has been discussed too much before https://rust-lang.zulipchat.com/#narrow/channel/146212-t-compiler.2Fconst-eval/topic/Const.20float.20math (feel free to drop by there if you have any interest working on it)

burrbull:
I'd wait for 1.83 (beta now): https://github.com/rust-lang/rust/pull/129195

kpreid:
> I think the better solution is to just move the math functions from std to core, and then make those const using the host machine's implementation.

Note that the Zulip discussion didn’t particularly reach a conclusion in favor of this plan, and in particular RalfJ opined that it is likely unsound:

> so... I think it is a hard constraint that all compile-time float support *must* be done via fully deterministic implementations that behave the same across all platforms

Also, applications may benefit from access to this same kind of determinism. For example, today I use `libm` not only for `no_std` compatibility, but also so that I can compute reproducible results. (Of course, future versions of `libm` may change algorithm and break those results, but dependency versions are controllable.)

Supporting `const` still might be declared infeasible for `libm` (as noted above about SIMD etc.); I just wanted to point out that _this particular alternate plan_ seems like it is not going to happen.

burrbull:
> Supporting `const` still might be declared infeasible for `libm` (as noted above about SIMD etc.); I just wanted to point out that _this particular alternate plan_ seems like it is not going to happen.

What you think about 2-level implementation?
- 1st level: constant naive fallible implementation.
```rust
pub const fn try_sin() -> Result<f32, Error> {};
```
[...]

PR #320, opened 2024-10-26
Add `copysignf16`, `copysignf128`, `fabsf16`, and `fabsf128`

tgross35:
Add generic versions of `abs` and `copysign`, which is used to implement `f16` and `f128` versions of these functions. These currently aren't tested because the builtin `musl` tests don't provide a reference, but the implementations are straightforward and we will be able to test against https://github.com/rust-lang/libm/pull/311 once it lands. (Technically musl on aarch64 has `fabsl` and `copysignl` as `f128`, but I don't think it is worth updating the serialization tests).

The second commit ("Add float and integer traits...") just copies the exact trait implementations that we use in `compiler-builtins`, with the assumption that we will use it more in the future and for testing.

The third commit ("Add f16 and f128 configuration...") copies the logic from `compiler-builtins` for setting `enable_f16` and `enable_f128` with the exception that here, the `unstable` feature must also be enabled. This will probably need to be tweaked a bit in the future so it isn't tied to other behavior enabled by `unstable`, but this is enough to get us started.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,13 +21,16 @@ arch = []
 
 # This tells the compiler to assume that a Nightly toolchain is being used and
 # that it should activate any useful Nightly things accordingly.
-unstable = ["unstable-intrinsics"]
+unstable = ["unstable-intrinsics", "unstable-float"]
 
 # Enable calls to functions in `core::intrinsics`
 unstable-intrinsics = []
 
 # Make some internal things public for testing.
 unstable-test-support = []
+
+# Enable the nightly-only `f16` and `f128`.
+unstable-float = []
 
 # Used to prevent using any intrinsics or arch-specific code.
 #
--- a/build.rs
+++ b/build.rs
@@ -1,6 +1,10 @@
 use std::env;
 
+mod configure;
+
 fn main() {
+    let cfg = configure::Config::from_env();
+
     println!("cargo:rerun-if-changed=build.rs");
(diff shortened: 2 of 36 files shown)


burrbull:
Using traits blocks us from using functions in `const` context, is not it?

tgross35:
It would indeed prevent us from making any of the public functions `const` (none of them are now), but providing `const` implementations is out of scope for this crate. Instead, there is work that can be done so that the equivalent functions in rust-lang/rust can become `const`, e.g. `const { 100.0f32.cos() }` will work.

I have some more details here if you are interested https://github.com/rust-lang/libm/issues/310

burrbull:
> is out of scope for this crate

Who has decided this?

tgross35:
It has been a goal for a while to make the float math functions available in `core`, which require various changes in this crate (improving accuracy, testing across all targets, performance, supporting the new float types). Conveniently, figuring out those same things will allow us to make the functions in `std`/`core` `const`. If that gets done then there won't be much use for this crate in its current form - so making that happen seems like a much better time investment.

The existing implementations aren't going away either. Generic methods just make it easier to support `f16` and `f128`, as well as `f32` algorithms that don't rely on `f64` (for cases where there is no hardfloat `f64`).

tgross35:
Superseded by a handful of PRs, including https://github.com/rust-lang/libm/pull/386.

PR #331, opened 2024-10-27
fix type of constants in ported sincosf

quaternic:
When porting sincosf from musl, the double-precision constants used for range reduction were mistakenly (I assume) cast to f32, which led to inaccurate results in some cases.

### DIFF
--- a/src/math/sincosf.rs
+++ b/src/math/sincosf.rs
@@ -131,14 +131,6 @@ pub fn sincosf(x: f32) -> (f32, f32) {
 #[cfg(test)]
 mod tests {
     use super::sincosf;
-    // use crate::_eqf;
-
-    // #[test]
-    // fn with_pi() {
-    //     let (s, c) = sincosf(core::f32::consts::PI);
-    //     _eqf(s.abs(), 0.0).unwrap();
-    //     _eqf(c, -1.0).unwrap();
-    // }
 
     #[test]
     fn rotational_symmetry() {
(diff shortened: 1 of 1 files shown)


src/math/sincosf.rs
```diff
@@ -128,14 +128,14 @@ pub fn sincosf(x: f32) -> (f32, f32) {
...
-    #[test]
-    fn with_pi() {
-        let (s, c) = sincosf(core::f32::consts::PI);
-        _eqf(s.abs(), 0.0).unwrap();
-        _eqf(c, -1.0).unwrap();
-    }
+    // use crate::_eqf;
+
+    // #[test]
+    // fn with_pi() {
```
tgross35, line 134: Does pi not pass anymore?

tgross35:
Thanks, does this lower the ULP error around multiples of pi/2?

src/math/sincosf.rs
quaternic: Because the patched function now produces more accurate results, (`sincosf(core::f32::consts::PI) = (-8.742278e-8, -1)` ), the old test doesn't pass. I'm not really sure what to do with it.

quaternic:
Yes. I've tested it on your `mpfr-test`-branch with the special casing for `sincosf` removed, and it seemed to pass everything. (Assuming I haven't missed some crucial thing and accidentally not tested it at all.)

src/math/sincosf.rs
tgross35: Oh, interesting. Maybe just delete the test and leave a `NB` comment for the module, in case we replace this with an implementation that handles it (or just special cases?)

tgross35:
Oh, awesome! Thanks for being thorough.

tgross35:
Thanks!

PR #335, opened 2024-10-28
Introduce an enum to represent different functions

tgross35:


tgross35:
Next steps so I don't forget:
- On the enum add functions for string names
- Make a single `mod acos { pub struct Func; }` within `func` for each function. This could just be a type, we use modules so we don't need to convert casing
- Make a trait for input and output types, name, base float type, etc. This should replace everything that the macro has to pass around all the time.
- Other traits can extend this, e.g. `MpOp` will say how to use Rug with that type

I think this should reduce how much we need to rely on proc macro magic

PR #348, opened 2024-10-29
Introduce generators that respect function domains

tgross35:
Introduce a `Domain` type for encoding function input range. This trait is used in `gen::domain` to create sequences of values that are either (1) around interesting points of this domain, or (2) logarithmically spaced within the domain.

Compared to the random generators, this means that we don't waste time checking large quantities of different NaNs or out of bound inputs (e.g. negative numbers and NaNs take up more than half the float space, this would be wasted checking `sqrt` which is only defined for x >= 0). It also means we know that coverage is uniform across the entire domain.

Currently only unary operations are supported.

This also includes a `f8` type that is just helpful for testing ULP ops since it is easily possible to list all values. I was going to remove this, but it turned out to be useful enough that I think I'll keep it around for future development.

### DIFF
--- a/crates/libm-test/src/domain.rs
+++ b/crates/libm-test/src/domain.rs
@@ -3,7 +3,7 @@
 use std::fmt;
 use std::ops::{self, Bound};
 
-use crate::Float;
+use crate::{Float, FloatExt};
 
 /// Representation of a function's domain.
 #[derive(Clone, Debug)]
@@ -19,7 +19,7 @@ pub struct Domain<T> {
 
 type BoxIter<T> = Box<dyn Iterator<Item = T>>;
 
-impl<F: Float> Domain<F> {
+impl<F: FloatExt> Domain<F> {
     /// The start of this domain, saturating at negative infinity.
     pub fn range_start(&self) -> F {
         match self.start {
--- a/crates/libm-test/src/domain.rs
+++ b/crates/libm-test/src/domain.rs
@@ -1,0 +1,186 @@
+//! Traits and operations related to bounds of a function.
+
+use std::fmt;
+use std::ops::{self, Bound};
+
+use crate::Float;
+
+/// Representation of a function's domain.
+#[derive(Clone, Debug)]
+pub struct Domain<T> {
+    /// Start of the region for which a function is defined (ignoring poles).
+    pub start: Bound<T>,
+    /// Endof the region for which a function is defined (ignoring poles).
+    pub end: Bound<T>,
+    /// Additional points to check closer around. These can be e.g. undefined asymptotes or
+    /// inflection points.
+    pub check_points: Option<fn() -> BoxIter<T>>,
+}
+
+type BoxIter<T> = Box<dyn Iterator<Item = T>>;
+
+impl<F: Float> Domain<F> {
+    /// The start of this domain, saturating at negative infinity.
+    pub fn range_start(&self) -> F {
+        match self.start {
+            Bound::Included(v) => v,
+            Bound::Excluded(v) => v.next_up(),
+            Bound::Unbounded => F::NEG_INFINITY,
+        }
+    }
+
+    /// The end of this domain, saturating at infinity.
+    pub fn range_end(&self) -> F {
+        match self.end {
+            Bound::Included(v) => v,
+            Bound::Excluded(v) => v.next_down(),
+            Bound::Unbounded => F::INFINITY,
+        }
+    }
+}
+
+impl<F: Float> Domain<F> {
+    /// x ∈ ℝ
+    pub const UNBOUNDED: Self =
+        Self { start: Bound::Unbounded, end: Bound::Unbounded, check_points: None };
+
+    /// x ∈ ℝ >= 0
+    pub const POSITIVE: Self =
+        Self { start: Bound::Included(F::ZERO), end: Bound::Unbounded, check_points: None };
+
+    /// x ∈ ℝ > 0
+    pub const STRICTLY_POSITIVE: Self =
+        Self { start: Bound::Excluded(F::ZERO), end: Bound::Unbounded, check_points: None };
+
+    /// Used for versions of `asin` and `acos`.
+    pub const INVERSE_TRIG_PERIODIC: Self = Self {
+        start: Bound::Included(F::NEG_ONE),
+        end: Bound::Included(F::ONE),
+        check_points: None,
+    };
+
+    /// Domain for `acosh`
+    pub const ACOSH: Self =
+        Self { start: Bound::Included(F::ONE), end: Bound::Unbounded, check_points: None };
+
+    /// Domain for `atanh`
+    pub const ATANH: Self = Self {
+        start: Bound::Excluded(F::NEG_ONE),
+        end: Bound::Excluded(F::ONE),
+        check_points: None,
+    };
+
+    /// Domain for `sin`, `cos`, and `tan`
+    pub const TRIG: Self = Self {
+        // TODO
+        check_points: Some(|| Box::new([-F::PI, -F::FRAC_PI_2, F::FRAC_PI_2, F::PI].into_iter())),
+        ..Self::UNBOUNDED
+    };
+
+    /// Domain for `log` in various bases
+    pub const LOG: Self = Self::STRICTLY_POSITIVE;
+
+    /// Domain for `log1p` i.e. `log(1 + x)`
+    pub const LOG1P: Self =
+        Self { start: Bound::Excluded(F::NEG_ONE), end: Bound::Unbounded, check_points: None };
+
+    /// Domain for `sqrt`
+    pub const SQRT: Self = Self::POSITIVE;
+
+    /// Domain for `gamma`
+    pub const GAMMA: Self = Self {
+        check_points: Some(|| {
+            // Negative integers are asymptotes
+            Box::new((0..u8::MAX).map(|scale| {
+                let mut base = F::ZERO;
+                for _ in 0..scale {
+                    base = base - F::ONE;
+                }
+                base
+            }))
+        }),
+        // Whether or not gamma is defined for negative numbers is implementation dependent
+        ..Self::UNBOUNDED
+    };
+
+    /// Domain for `loggamma`
+    pub const LGAMMA: Self = Self::STRICTLY_POSITIVE;
+}
+
+/// Implement on `op::*` types to indicate how they are bounded.
+pub trait HasDomain<T>
+where
+    T: Copy + fmt::Debug + ops::Add<Output = T> + ops::Sub<Output = T> + PartialOrd + 'static,
+{
+    const DOMAIN: Domain<T>;
+}
+
+/// Implement [`HasDomain`] for both the `f32` and `f64` variants of a function.
+macro_rules! impl_has_domain {
+    ($($fn_name:ident => $domain:expr;)*) => {
+        paste::paste! {
+            $(
+                // Implement for f64 functions
+                impl HasDomain<f64> for $crate::op::$fn_name::Routine {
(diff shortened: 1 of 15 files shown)


tgross35:
Rebased to build on https://github.com/rust-lang/libm/pull/349

tgross35:
I think the current logic is pretty good. Still needs cleanup but these are the check points in debug mode:

sqrt-inputs
cbrt-inputs
cos-inputs

Release numbers take forever to generate the plots but it just shows more even coverage as expected

sqrt-inputs-log

tgross35:
@beetrees @quaternic you both suggested this at different points, would you mind reviewing this? I still have to wire up tests but I think the generator itself is good, see the above plots.

crates/libm-test/src/domain.rs
```diff
@@ -0,0 +1,213 @@
...
+/// implement `HasDomain`.
+pub trait Domain<T>
+where
+    T: Copy + fmt::Debug + ops::Add<Output = T> + ops::Sub<Output = T> + PartialOrd + 'static,
+{
+    /// The region for which the function is defined. Ignores poles.
+    const DEFINED: (Bound<T>, Bound<T>);
+
+    /// The region, if any, for which the function repeats. Used to test within.
+    const PERIODIC: Option<(Bound<T>, Bound<T>)> = None;
```
quaternic, line 21: I'm not convinced periodicity will be that useful here, since not that many of the functions actually are periodic. Mainly `sin`, `cos`, `tan`, and `fract` I suppose. Out of those, only `fract` has a period (`1.0`) which is actually representable as a float. The functions periodic in `π` are problematic exactly because of that; the worst-case inputs are very large ones where the function just happens to land very close to zero. Unlike `fract` which is just exactly zero for numbers large enough.

tgross35: I got this from another math library (I thought it was core-math maybe not since I can't find it now...) that suggested the heaviest testing within one period to get better coverage with the larger float types. But in retrospect this doesn't make a lot of sense to me given that -π..π already includes more than 50% of the float's representable values.

Do you think it is better to just drop any references to periodicity and instead test with `Unbounded`? Or is there still some advantage to heavier testing within this range, or keeping the information around for some future use.

quaternic: I'd just drop it, but it's not something I've thought about all that much. It could be useful to have the information somewhere (e.g. [0,2π] or [-π,π] might be very reasonable input ranges to have on some benchmarks) but it's mostly only relevant for sin/cos/tan.

crates/libm-test/src/gen/domain.rs
```diff
@@ -0,0 +1,207 @@
...
+            5_000_000
+        }
+    } else {
+        // Without optimizations just run a quick check
+        800
+    }
+};
+
+/// Some functions have infinite asymptotes, limit how many we check.
+const MAX_ASYMPTOTES: usize = 10;
```
tgross35, line 44: I need to refactor this all a bit considering I just copied and pasted NTESTS from the random tests. Still brainstorming how to do this (input appreciated) but I'm thinking maybe something like:

- Have `N` number of tests by default, which is 5M in the existing code (but probably needs to be reduced based on the below)
- `f32` unary functions run `N` random tests
- `f64` and `f128` unary functions run `N * 4` random tests
- If the function takes two inputs, multiply the number of tests by 4. If it takes three inputs, multiply by 8.
- If a domain test exists for the function, run `N` domain-based tests and reduce the number of random tests by a factor of 100
- If an exhaustive test for `f32` (not yet implemented) or high-iteration test (`f64`/`f128` or multi-input functions, also not implemented) should be run, still run the "interesting points" portion of domain-based tests but replace the `logspace` tests with whatever fits. I should probably split this into two generators rather than chaining...

Basically in all cases, either the exhaustive tests or the logspace are going to consume the bulk of time. But we still want to run "interesting points" and the random tests in hopes that they will find errors earlier than waiting for the whole exhaustive check to run. Also random tests should cover some signaling NaNs.

crates/libm-test/src/domain.rs
tgross35: Yeah, it applies to a lot less than I expected it to when starting this. Okay, I'll remove it (but not immediately in case other refactoring comes up in review).

image

Github thinks you're going to comment in the future? 🤔

quaternic: I was wondering about that as well, good to see it's just a visual glitch. Yours is too 🤔 Apparently the US is turning clocks back at this time (daylight savings)

src/math/support/float_traits.rs
```diff
@@ -69,7 +78,19 @@ pub trait Float:
...
+            // }
+            // fn is_nan(x: Self) -> bool {
+            // When using mangled-names, the "real" compiler-builtins might not have the
+            // necessary builtin (__unordtf2) to test whether `f128` is NaN.
+            // FIXME(f16_f128): Remove once the nightly toolchain has the __unordtf2 builtin
+            // x is NaN if all the bits of the exponent are set and the significand is non-0
+            x.to_bits() & Self::EXP_MASK == Self::EXP_MASK
+                && x.to_bits() & Self::SIG_MASK != Self::Int::ZERO
+        };
+        if is_nan(self) && is_nan(rhs) { true } else { self.to_bits() == rhs.to_bits() }
```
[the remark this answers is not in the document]
tgross35: Oh right, we could do this in builtins as well.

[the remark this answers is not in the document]
tgross35: Applied here, thanks.

crates/libm-test/src/f8_impl.rs
```diff
@@ -0,0 +1,489 @@
...
+
+    fn normalize(_significand: Self::Int) -> (i32, Self::Int) {
+        unimplemented!()
+    }
+}
+
+impl f8 {
+    pub const ALL_LEN: usize = 240;
+
+    /// All non-infinite non-NaN values of `f8` excluding `-0`.
```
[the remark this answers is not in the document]
tgross35: Removed this.

crates/libm-test/src/gen/domain.rs
```diff
@@ -0,0 +1,207 @@
...
+    for x in [F::ONE, F::NEG_ONE, F::ZERO] {
+        around(x, &mut values);
+    }
+
+    // Values around min and max are interesting
+    values.extend(count_up(F::MIN).take(AROUND));
+    values.extend(count_down(F::MAX).take(AROUND));
+
+    // Check some special values that aren't included in the above ranges
+    values.push(F::NAN);
```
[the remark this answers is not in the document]
tgross35: Good thought, I added a handful of relevant https://github.com/rust-lang/libm/blob/037a7d66848b2fd78c2c10eabe76a3dd9d529cb9/crates/libm-test/src/num.rs#L164-L173.

These edge case tests will probably get split to a separate file.

crates/libm-test/src/domain.rs
tgross35: Alright, I removed `PERIODIC` and the logic relating to it.

crates/libm-test/src/f8_impl.rs
```diff
@@ -0,0 +1,489 @@
...
+    pub const ALL: [Self; Self::ALL_LEN] = [
```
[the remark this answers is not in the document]
tgross35: (Thought I replied before) it definitely could be simplified, I just like having having the table available as a quick reference.

crates/libm-test/src/num.rs
```diff
@@ -0,0 +1,360 @@
...
+/// to logarithmic spacing of their values.
+///
+/// Note that this tends to skip negative zero, so that needs to be checked explicitly.
+pub fn logspace<F: FloatExt>(start: F, end: F, steps: F::Int) -> impl Iterator<Item = F> {
+    assert!(!start.is_nan());
+    assert!(!end.is_nan());
+    assert!(end >= start);
+
+    let mut steps = steps.checked_sub(F::Int::ONE).expect("`steps` must be at least 2");
+    let between = start.ulp_between(end).expect("`start `must be less than `end`");
```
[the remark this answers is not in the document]
tgross35: Good catch, and there is a check for `end >= start` above. I updated this message.

crates/libm-test/src/gen/domain.rs
tgross35: I have this as part of https://github.com/rust-lang/libm/pull/364.

crates/libm-test/src/num.rs
```diff
@@ -0,0 +1,447 @@
...
+
+    let x_abs = x.abs().to_bits();
+    let y_abs = y.abs().to_bits();
+
+    if x.signum() == y.signum() {
+        Some(max(x_abs, y_abs) - min(x_abs, y_abs))
+    } else {
+        Some(x_abs + y_abs)
+    }
+}
```
quaternic, line 202: Could extract a helper, which might also be useful for some of the other functions in this module:
```suggestion
pub fn ulp_between<F: Float>(x: F, y: F) -> Option<F::Int> {
    let a = as_ulp_steps(x)?;
    let b = as_ulp_steps(y)?;

    Some(a.abs_diff(b))
}
/// Return the (signed) number of steps from zero to x
fn as_ulp_steps(x: f32) -> Option<F::SignedInt> {
    let s = x.to_bits_signed();

    let val = if s >= 0 {
        // each increment from `s = 0`
        // is one step up from `x = 0.0`
        s
    } else {
        // each increment from `s = F::SignedInt::MIN`
        // is one step down from `x = -0.0`
        F::SignedInt::MIN - s
    };

    if (s as F::Int & !F::SIGN_MASK) <= F::EXP_MASK {
        Some(val)
    } else {
        // x is NaN
        None
    }
}
```

tgross35: This wasn't causing the tests to fail and I couldn't figure out why, turns out you found a bug in `abs_diff` :) https://github.com/rust-lang/libm/pull/366. Updated with a small adjustment from your suggestion.

crates/libm-test/src/gen/edge_cases.rs
```diff
@@ -0,0 +1,88 @@
...
+    if let Some(f) = domain.check_points {
+        let iter = f();
+        for x in iter.take(MAX_CHECK_POINTS) {
+            count_up(x, values);
+            count_down(x, values);
+        }
+    }
+
+    // Some results may overlap so deduplicate the vector to save test cycles.
+    values.dedup_by_key(|x| x.to_bits());
```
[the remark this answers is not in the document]
tgross35: Ah, I had these sorted at one point but must have lost it in refactoring. Added it back.

crates/libm-test/src/domain.rs
```diff
@@ -0,0 +1,184 @@
...
+    pub check_points: Option<fn() -> BoxIter<T>>,
+}
+
+type BoxIter<T> = Box<dyn Iterator<Item = T>>;
+
+impl<F: Float> Domain<F> {
+    /// The start of this domain, saturating at negative infinity.
+    pub fn range_start(&self) -> F {
+        match self.start {
+            Bound::Included(v) | Bound::Excluded(v) => v,
```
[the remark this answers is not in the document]
tgross35: I did this intentionally since it seems the bound is probably worth checking even if it is excluded, but yeah this isn't technically correct. Any idea how to express this better?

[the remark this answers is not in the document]
tgross35: Good call, added this. (Also removed the `is_infinite` checks within `count_up`/`count_down` since they don't serve much purpose and allow this to be cleaner).

tgross35:
I think this is in a pretty reasonable state so I am going to go ahead and merge. As always, further reviews are welcome.

PR #349, opened 2024-10-30
Migrate types from a macro to a trait

tgross35:
This allows us to move a lot of logic out of macros and into generic functions, which is a nice cleanup and much better for usability.

### DIFF
--- a/crates/libm-macros/Cargo.toml
+++ b/crates/libm-macros/Cargo.toml
@@ -8,6 +8,7 @@ publish = false
 proc-macro = true
 
 [dependencies]
+heck = "0.5.0"
 proc-macro2 = "1.0.88"
 quote = "1.0.37"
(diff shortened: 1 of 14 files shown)


tgross35:
@beetrees I'm still working on some better naming cleanup / docs / comments, but you may want to take a look at this. Basically migrating from always relying on macro expansion to putting everything into a trait, which should help avoid string matching problems (like you pointed out on a different PR).

None of the possible refactoring of ULP / precision is done here.

tgross35:
Let me know if you have any name suggestions too. I feel like too much here is some variation of "function" or "op", but I can't come up with anything much better.

tgross35:
I still don't have great names but I added the following to hopefully disambiguate a bit https://github.com/rust-lang/libm/blob/7c418306c90bf0180249b3f38761e416cfee86a9/crates/libm-test/src/op.rs#L6-L14

PR #351, opened 2024-10-31
Introduce `hf32!` and `hf64!` macros for hex float support

tgross35:
Rust does not have any native way to parse hex floats, but they are heavily used in the C algorithms that we derive from. Introduce a const function that can parse these, as well as macros `hf32!` and `hf64!` that ensure the string literals get handled at compiler time.

These are currently not used but making everything available now will ease future development.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -44,37 +44,19 @@ Check [PR #65] for an example.
   `mod.rs`.
 
 - You may encounter weird literals like `0x1p127f` in the MUSL code. These are hexadecimal floating
-  point literals. Rust (the language) doesn't support these kind of literals. The best way I have
-  found to deal with these literals is to turn them into their integer representation using the
-  [`hexf!`] macro and then turn them back into floats. See below:
+  point literals. Rust (the language) doesn't support these kind of literals. This crate provides
+  two macros, `hf32!` and `hf64!`, which convert string literals to floats at compile time.
 
-[`hexf!`]: https://crates.io/crates/hexf
-
-``` rust
-// Step 1: write a program to convert the float into its integer representation
-#[macro_use]
-extern crate hexf;
-
-fn main() {
-    println!("{:#x}", hexf32!("0x1.0p127").to_bits());
-}
-```
-
-``` console
-$ # Step 2: run the program
-$ cargo run
-0x7f000000
-```
-
-``` rust
-// Step 3: copy paste the output into libm
-let x1p127 = f32::from_bits(0x7f000000); // 0x1p127f === 2 ^ 12
-```
+  ```rust
+  assert_eq!(hf32!("0x1.ffep+8").to_bits(), 0x43fff000);
+  assert_eq!(hf64!("0x1.ffep+8").to_bits(), 0x407ffe0000000000);
+  ```
 
 - Rust code panics on arithmetic overflows when not optimized. You may need to use the [`Wrapping`]
-  newtype to avoid this problem.
+  newtype to avoid this problem, or individual methods like [`wrapping_add`].
 
 [`Wrapping`]: https://doc.rust-lang.org/std/num/struct.Wrapping.html
+[`wrapping_add`]: https://doc.rust-lang.org/std/primitive.u32.html#method.wrapping_add
 
 ## Testing
 
--- a/src/math/support/hex_float.rs
+++ b/src/math/support/hex_float.rs
@@ -1,0 +1,399 @@
+//! Utilities for working with hex float formats.
+
+#![allow(dead_code)] // FIXME: remove once this gets used
+
+/// Construct a 32-bit float from hex float representation (C-style)
+pub const fn hf32(s: &str) -> f32 {
+    f32_from_bits(parse_any(s, 32, 23) as u32)
+}
+
+/// Construct a 64-bit float from hex float representation (C-style)
+pub const fn hf64(s: &str) -> f64 {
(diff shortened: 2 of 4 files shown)


tgross35:
@quaternic since you have already looked at this algorithm some, would you mind reviewing it here? :)

quaternic:
On a high level, I have two considerations:

Firstly, I think the parsing code could be streamlined significantly. As one particular idea, the `parse_any` could return something like `(M, E): (i128, i32)` with numeric value `M * 2^E`. This is more general, and would separate (parsing an abstract float) from (constructing a value of `fN`).

Secondly, it should handle inputs that cannot be represented in the target type, alternatives:
 - report an error (preferred)
 - correct rounding
 
Allowing for both might be ideal, but it's probably not needed for the purposes of this crate. The bug in `sincosf` was in part due to the compiler accepting `const FOO: f32 = /*literal with lots of digits*/`. In decimal you *need* `0.1` to get rounded, but with hex you can always express the rounded value with fewer digits.

tgross35:
Thanks for the feedback! I updated with your suggestion for handling excess precision, that could use a look.

I am not sure what to 
(I have no clue why this would be getting a SIGILL on ppc but I am sure it is unrelated - maybe something with how `should_panic` is trying to catch unwinds)

CONTRIBUTING.md
```diff
@@ -44,37 +44,19 @@ Check [PR #65] for an example.
   `mod.rs`.
 
 - You may encounter weird literals like `0x1p127f` in the MUSL code. These are hexadecimal floating
-  point literals. Rust (the language) doesn't support these kind of literals. The best way I have
-  found to deal with these literals is to turn them into their integer representation using the
-  [`hexf!`] macro and then turn them back into floats. See below:
+  point literals. Rust (the language) doesn't support these kind of literals. This crate provides
+  two macros, `hf32!` and `hf64`, which convert string literals to floats at compile time.
```
fbstj, line 48:
```suggestion
  two macros, `hf32!` and `hf64!`, which convert string literals to floats at compile time.
```

CONTRIBUTING.md
```diff
@@ -44,37 +44,19 @@ Check [PR #65] for an example.
...
-$ cargo run
-0x7f000000
-```
-
-``` rust
-// Step 3: copy paste the output into libm
-let x1p127 = f32::from_bits(0x7f000000); // 0x1p127f === 2 ^ 12
-```
+  ```rust
+  assert_eq!(hf64!("0x1.ffep+8").to_bits(), 0x43fff000);
```
fbstj, line 51: is this meant to be a `hf32!` example?

tgross35: Ah, good catch!

quaternic:
> I am not quite sure what you have in mind re. returning `(i128, i32)` because `parse_any` always needs `sig_bits` to figure out exponent bounds and do the shifting, so it seems reasonable to just clear the implicit bit and reconstruct the float there rather than reconstructing everywhere it is called. Unless you just mean to factor some of that out to a new function?

Right, yes, I meant refactoring it to separate parsing from the details of the target representation.

>(feel free to sketch it out on the playground if you had something in mind - I'm not tied to this implementation, it was just something I could get together quick)

Alright, I can do that!

quaternic:
Why, oh why, did I volunteer to write parsing code. It certainly wasn't as quick and easy as suggesting such refactoring was. It's not really a sketch anymore, but pretty much a full rewrite of the internals: [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=90a7994e42e22b18b87ea32c4edf36e3)

In functionality, one concrete improvement is support for any integer parts as in `0x123.4p-56`. I followed whatever IEEE says for the parsed format, so it accepts some additional variations, such as `0X` as the prefix, or no leading zero as in `0x.123p0`.

tgross35:
> Why, oh why, did I volunteer to write parsing code. (edit: As a `const fn`, even!) It certainly wasn't as quick and easy as suggesting such refactoring was. It's not really a sketch anymore, but pretty much a full rewrite of the internals: [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=90a7994e42e22b18b87ea32c4edf36e3)

Well, thanks for doing it! `const fn` for when you miss C style array logic... The code is certainly much cleaner than what I hacked together :) 

> In functionality, one concrete improvement is support for any integer parts as in `0x123.4p-56`. I followed whatever IEEE says for the parsed format, so it accepts some additional variations, such as `0X` as the prefix, or no leading zero as in `0x.123p0`.

That wasn't even a goal since C never produces those, above and beyond!

I had to massage a few things but I basically picked up exactly what you had. I'll merge it in a bit unless anything else jumps out to you.

src/math/support/hex_float.rs
```diff
@@ -0,0 +1,399 @@
...
+    fn test_parse_any() {
+        for k in -149..=127 {
+            let s = format!("0x1p{k}");
+            let x = hf32(&s);
+            let y = if k < 0 { 0.5f32.powi(-k) } else { 2.0f32.powi(k) };
+            assert_eq!(x, y);
+        }
+
+        let mut s = *b"0x.0000000p-121";
+        for e in 0..10 {
```
quaternic, line 195:
```suggestion
        for e in 0..40 {
```
Makes the test more comprehensive by including the transition from subnormals to normals.

quaternic:
> I had to massage a few things but I basically picked up exactly what you had. I'll merge it in a bit unless anything else jumps out to you.

Thanks for putting it together! Looks about right to me.

Issue #336, opened 2024-10-28
Would adding docstrings to functions be a welcome PR?


JSorngard:
I notice that a lot of the functions don't have a docstring that shows up on `docs.rs`, but there is information available in comments in the implementation. Would it be welcome if I made a PR where I add docstrings to some more functions?

JSorngard:
And if so, what should they contain beyond a basic description of the function? Error bounds?

tgross35:
Sure! A short description is totally fine, maybe with a link to the wikipedia page for anything less trivial.

tgross35:
I am going to close this since a couple PRs were merged. Further documentation is always welcome :)

PR #352, opened 2024-10-31
Add some more basic docstrings

JSorngard:
Adds docstrings to `tgamma`, `tgammaf`, `lgamma`, `lgammaf`, `trunc`, `truncf`, `exp10`, and `exp10f`. Related to #336.

### DIFF
--- a/src/math/lgamma.rs
+++ b/src/math/lgamma.rs
@@ -1,5 +1,6 @@
 use super::lgamma_r;
 
+/// The natural logarithm of the [Gamma function](https://en.wikipedia.org/wiki/Gamma_function) (f64).
 #[cfg_attr(all(test, assert_no_panic), no_panic::no_panic)]
 pub fn lgamma(x: f64) -> f64 {
     lgamma_r(x).0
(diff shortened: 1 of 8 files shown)


JSorngard:
Should I specify that `trunc` rounds to the *nearest* integer towards 0?

src/math/lgamma.rs
```diff
@@ -1,5 +1,6 @@
 use super::lgamma_r;
 
+/// The natural logarithm of the [Gamma function](https://en.wikipedia.org/wiki/Gamma_function) (f64).
```
tgross35, line 3: Minor nit since this line is a couple characters over 100 (same for f32)

```suggestion
/// The natural logarithm of the [Gamma function] (f64).
///
/// [Gamma function] https://en.wikipedia.org/wiki/Gamma_function
```

tgross35:
> Should I specify that `trunc` rounds to the _nearest_ integer towards 0? I currently just tried to stay as faithful as I could to the documentation of `musl`'s trunc function.

Whatever you think would be helpful reading it :) there is no need to keep the wording from specifically musl, sources like e.g. [cppreference](https://en.cppreference.com/w/c/numeric/math/trunc) likely have more thorough docs.

JSorngard:
Alright! I wrote it as clearly as I could and added a linebreak to the lgamma docstrings.

tgross35 (approved):
Thanks!

PR #353, opened 2024-10-31
ci: add support for loongarch64-unknown-linux-gnu

heiher:
Add support for `loongarch64-unknown-linux-gnu` target in CI.


tgross35 (approved):
Thanks!

PR #365, opened 2024-12-22
Remove tests against system musl

tgross35:
We now have tests against our custom-built musl as well as tests against MPFR. The tests against system musl covers less than those against custom-built musl, and are less portable; there isn't much benefit to keeping them around so just remove them.


tgross35:
Let's clean it up 🧹

PR #364, opened 2024-12-21
Add exhaustive/extensive tests

tgross35:
Add a generator that will test all inputs for input spaces `u32::MAX` or smaller (e.g. single-argument `f32` routines).  For anything larger, still run approximately `u32::MAX` tests, but distribute inputs evenly across the function domain.

Since we often only want to run one of these tests at a time, this implementation parallelizes within each test using `rayon`. A custom test runner is used so a progress bar is possible.

Specific tests must be enabled by setting the `LIBM_EXTENSIVE_TESTS` environment variable, e.g.

    LIBM_EXTENSIVE_TESTS=all_f16,cos,cosf cargo run ...

Testing on a recent machine, most tests take about two minutes or less. The Bessel functions are quite slow and take closer to 10 minutes, and FMA is increased to run for about the same.

PR #379, opened 2024-12-29
Streamline the way that test iteration count is determined  (replace `NTESTS`)

tgross35:
Currently, tests use a handful of constants to determine how many iterations to perform: `NTESTS`, `AROUND`, and `MAX_CHECK_POINTS`. This configuration is not very straightforward to adjust and needs to be repeated everywhere it is used.

Replace this with new functions in the `run_cfg` module that determine iteration counts in a more reusable and documented way.

This also changes the random generator to generate inputs on the fly rather than caching.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -113,6 +113,11 @@ jobs:
         rustup target add x86_64-unknown-linux-musl
         cargo generate-lockfile && ./ci/run-docker.sh ${{ matrix.target }}
 
+    - name: Print test logs if available
+      if: always()
+      run: if [ -f "target/test-log.txt" ]; then cat target/test-log.txt; fi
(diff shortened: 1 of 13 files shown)


tgross35:
@beetrees if you get the chance, could you give this a second set of eyes? Mostly just the rewrite of `random.rs` and that the logic in `run_cfg` looks reasonable.

I changed the PR for extensive tests to also build off of this, I'll get that one ready next https://github.com/rust-lang/libm/pull/364.

tgross35:
Added @quaternic's Zulip suggestion to randomize the seed with a way to override it so we wind up with better test coverage over time.

tgross35:
I am going to merge this so I can get other dependent changes in, as always continued reviews are welcome.

PR #391, opened 2025-01-03
Add `ldexpf16`, `ldexpf128`, `scalbnf16`, and `scalbnf128`

tgross35:
The `scalbn` functions are similar enough that they can easily be made generic. Do so and add `f16` and `f128` versions.


tgross35:
`f16` and `f128` versions are untested, blocked until I can figure out how to do that with MPFR.

tgross35:
I think the f16 version is running into the same issue as https://github.com/rust-lang/libm/pull/419#issuecomment-2608753669 (among other issues) so I am just going to drop that for now, keeping only scalbnf128.

tgross35:
Oh, interesting; the same problem seems to be showing up for the `f128` version:

```
thread 'mp_edge_case_scalbnf128' panicked at crates/libm-test/tests/multiprecision.rs:17:48:
called `Result::unwrap()` on an `Err` value:
    input:    (0x00000000000000000000000000000003, -2) (0x00000000000000000000000000000003, 0xfffffffe)
    expected: 0x00000000000000000000000000000000 0x00000000000000000000000000000000
    actual:   0x00000000000000000000000000000001 0x00000000000000000000000000000001

Caused by:
    ulp 1 > 0
```

Testing with the new (unreleased) version of Rug returns 0x00000000000000000000000000000001, so I will just wait for that release.

PR #392, opened 2025-01-03
Add tests against MPFR for `scalbn{f}` and `ldexp{f}`

tgross35:


crates/libm-test/src/mpfloat.rs
```diff
@@ -255,6 +255,36 @@ macro_rules! impl_op_for_ty {
...
+                    (new_mpfloat::<Self::FTy>(), new_mpfloat::<Self::FTy>())
+                }
+
+                fn run(this: &mut Self::MpTy, input: Self::RustArgs) -> Self::RustRet {
+                    this.0.assign(input.0);
+                    this.1.assign(2.0);
+                    this.1.pow_assign(input.1);
+                    let ord = this.0.mul_assign_round(&this.1, Nearest);
+                    prep_retval::<Self::FTy>(&mut this.0, ord)
+                }
```
tgross35, line 285: This discards the rounding of `2 ^ exp` which should be okay since powers of two are representable if within range. Somebody should double check this though.

tgross35: Ah this doesn't work because `scalbn(INF, -HUGE)` works out to `INF * 0 = NaN`, but the result should be infinity.

programmerjake: after checking the docs, what you want is [`Float::shl`](https://docs.rs/rug/latest/rug/struct.Float.html#impl-Shl%3Ci32%3E-for-Float), if you're not immediately converting back to a primitive float you'll probably also want to call [`subnormalize`](https://docs.rs/rug/latest/rug/struct.Float.html#method.subnormalize)

programmerjake: you can look at `subnormalize_ieee_round`'s source for an idea of how to use `subnormalize`, i'm assuming rust may gain support for non-ieee floats such as bf16 or f80.

tgross35: Thanks, that does seem to work simpler than what I had. The call to `prep_retval` does call `subnormalize_ieee_round` before casting to the float type ([link](https://github.com/rust-lang/libm/blob/1b8940071483790bfb311f7ffacba0eb4186cc84/crates/libm-test/src/mpfloat.rs#L22-L28)), I don't think `bf16` uses `libm` (unlike `f16`, all ops may be done as f32 right?) but this will probably need to be adjusted once we add the x87 float type. Easiest to just cross that bridge later though, I'm just getting started with `f16`/`f128` here.

PR #400, opened 2025-01-05
Add tests against MPFR for `remquo` and `remquof`

tgross35:
Rug does not yet expose this function, but it is possible to use the MPFR bindings directly.


tgross35:
Currently the top bit disagrees with MPFR some of the time. I am not sure if I just need to truncate differently or if this is a result of the `TODO: this matches musl behavior, but it is incorrect` comment in source.

PR #404, opened 2025-01-06
Enable MPFR tests on i586

tgross35:


programmerjake:
i think for `i586` it should just use `frndint` with setting the x87 CW (actually not slow) for `floor`/`ceil`. like: https://github.com/NetBSD/src/blob/20433927938987dd64c8f6aa46904b7aca3fa39e/lib/libm/arch/i387/s_ceil.S

if `asm!` supported `x87` registers as inputs/outputs, that would be nice...

tgross35:
> i think for `i586` it should just use `frndint` with setting the x87 CW (actually not slow) for `floor`/`ceil`. like: https://github.com/NetBSD/src/blob/20433927938987dd64c8f6aa46904b7aca3fa39e/lib/libm/arch/i387/s_ceil.S

These showed up when I (recently) enabled tests against musl for edge cases, I just marked them xfail to unblock adding tests 5e0c8403ae508465cfacce8ba96458e00f28a753. This sounds like a good idea though, assuming that is license-compatible (looks like it) we could probably reuse a few x87 routines from there. Opened https://github.com/rust-lang/libm/issues/407 to track it.

Somewhat more concerning is that it looks like our `fmaf` and `fdim` are not always accurate on this platform...

> if `asm!` supported `x87` registers as inputs/outputs, that would be nice...

Yeah, I know a `f80` in `core::arch` has been discussed a lot. I'll probably pick this up after `f128` and `f16` get further along if nobody else has gotten to it at that point.

programmerjake:
> > if `asm!` supported `x87` registers as inputs/outputs, that would be nice...
> 
> Yeah, I know a `f80` in `core::arch` has been discussed a lot. I'll probably pick this up after `f128` and `f16` get further along if nobody else has gotten to it at that point.

i meant passing a `f32` or `f64` in a x87 register, we don't need to wait for `f80` for that.

PR #388, opened 2025-01-02
Add exhaustive/extensive tests

tgross35:
Add a generator that will test all inputs for input spaces `u32::MAX` or smaller (e.g. single-argument `f32` routines).  For anything larger, still run approximately `u32::MAX` tests, but distribute inputs evenly across the function domain.

Since we often only want to run one of these tests at a time, this implementation parallelizes within each test using `rayon`. A custom test runner is used so a progress bar is possible.

Specific tests must be enabled by setting the `LIBM_EXTENSIVE_TESTS` environment variable, e.g.

    LIBM_EXTENSIVE_TESTS=all_f16,cos,cosf cargo run --features test-multiprecision --release

Testing on a recent machine, most tests take about two minutes or less. The Bessel functions are quite slow and take closer to 10 minutes, and FMA is increased to run for about the same.

This uses the `libtest_mimc` harness so a progress bar can be printed. Example:

```
running 90 tests
test mp_extensive_acos       ... ignored, extensive tests are only run if specified in RUST_LIBM_EXTENSIVE
test mp_extensive_acosf      ... starting extensive tests for `acosf`
[9m  100%] #################### acosf 4,294,967,295/4,294,967,295: 7,672,836.6102/s eta 0s
ok
test mp_extensive_acosh      ... ignored, extensive tests are only run if specified in RUST_LIBM_EXTENSIVE
test mp_extensive_acoshf     ... starting extensive tests for `acoshf`
[5m  73 %] ###############----- acoshf 3,139,620,000/4,294,967,295: 3,544,521.3401/s eta 5m (1.6715916e36,)
```

Recreated from https://github.com/rust-lang/libm/pull/364.

PR #405, opened 2025-01-06
Add more detailed definition output for `update-api-list.py`

tgross35:
Update the script to produce, in addition to the simple text list, a JSON file listing routine names, the types they work with, and the source files that contain a function with the routine name. This gets consumed by another script and will be used to determine which extensive CI jobs to run.


tgross35:
Cherry picked from https://github.com/rust-lang/libm/pull/388.

PR #416, opened 2025-01-07
Make extensive tests exhaustive if there are enough iterations available

### LINKED ISSUE #409: Extensive tests for integers need to use intervals
Integer inputs don't currently respect the iteration count, meaning that e.g tests for `ldexp` will run forever.

beetrees:
Currently the extensive tests will only check values within a functions domain (or that are non-NaN for multi-argument functions), even when there are enough iterations available to check every single possible value. This PR fixes that. There was also a bug where trying to exhaustively test `fma`/`fmaf` on 32-bit targets would immediately panic as the test case count doesn't fit in a 32-bit `usize`: this PR changes `get_test_cases` to use a separate `u64` instead of `ExactSizeIterator` to fix this.

This also fixes #409.

### DIFF
--- a/crates/libm-test/src/gen/domain_logspace.rs
+++ b/crates/libm-test/src/gen/domain_logspace.rs
@@ -27,5 +27,5 @@ where
     let start = domain.range_start();
     let end = domain.range_end();
     let steps = OpITy::<Op>::try_from(ntests).unwrap_or(OpITy::<Op>::MAX);
-    logspace(start, end, steps).map(|v| (v,))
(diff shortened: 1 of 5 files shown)


tgross35:
Just as a note I'm currently planning to rework some of this so that the `domain_logspace` and `extensive` generators can use the same code, but that is pretty early on still https://github.com/rust-lang/libm/pull/415.

Also if you're currently running this locally you need to skip the `jn*` and `yn*` tests, MPFR often throws an allocation failure that I haven't yet looked into.

PR #422, opened 2025-01-11
Reorder tests in `run.sh`

tgross35:
I do not believe Cargo separately caches crates with different sets of features enabled. So, ensuring that tests run with `unstable-intrinsics` are always grouped should slightly reduce runtime.

As an added benefit, all the debug mode tests run first so initial feedback is available faster.


tgross35:
It seems like there may be some minor improvements compared to the previous run.

PR #424, opened 2025-01-12
Add a `release-checked` profile with debug and overflow assertions

tgross35:
A failing debug assertion or overflow without correctly wrapping or saturating is a bug, but the `debug` profile that has these enabled does not run enough test cases to hit edge cases that may trigger these. Add a new `release-checked` profile that enables debug assertions and overflow checks. This seems to only extend per-function test time by a few seconds (or around a minute on longer extensive tests), so enable this as the default on CI.

In order to ensure `no_panic` still gets checked, add a build-only step to CI.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -238,7 +238,8 @@ jobs:
 
           LIBM_EXTENSIVE_TESTS="$CHANGED" cargo t \
             --features test-multiprecision,unstable \
-            --release -- extensive
+            --profile release-checked \
+            -- extensive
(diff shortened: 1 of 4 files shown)


tgross35:
This adds a couple of minutes to the longest CI job, effectively undoing wins from https://github.com/rust-lang/libm/pull/422. Oh well, I think that is a reasonable price to pay to make sure we don't OOB our lookup tables (hopefully that would cause an error anyway, but it doesn't hurt to be sure).

PR #430, opened 2025-01-12
wasm32 and aarch64 intrinsics for rint and rintf

### LINKED ISSUE #421: Use `core::arch` intrinsics for for `rint` and `rintf` where possible
These functions don't have any arch-specialized versions right now. I believe they can be done with smaller code size and better performance on some targets:

- On Wasm, `f{32,64}_nearest` are always available and already exposed in core::arch (cc #418).
- AArch64 with target_feature=neon has `frintn` but the scalar versions don't seem to be exposed in core::arch. I guess the vector versions (e.g., `vrndn_f32`) are still an improvement over softfloat even if they need an unnecessary scalar->vector splat.
-  On x86 and x86_64, there's `_mm_round_{sd,ss}` and `_MM_FROUND_TO_NEAREST_INT`. But this requires SSE4.1 so it's not applicable unless someone happens to build libm from crates.io with appropriate RUSTFLAGS.
    - [GCC manages to emit inline code for it with just SSE2](https://godbolt.org/z/jff9r88ev) but it's not exactly pretty and I'm not sure if it's even significantly better than the generic implementation in this crate.

hanna-kruppe:
I didn't bother with x86 because SSE4.1-exclusive code paths are extra testing burden that's unlikely to benefit many people.

Closes #421

### DIFF
--- a/src/math/arch/aarch64.rs
+++ b/src/math/arch/aarch64.rs
@@ -1,0 +1,33 @@
+use core::arch::aarch64::{
+    float32x2_t, float64x1_t, vdup_n_f32, vdup_n_f64, vget_lane_f32, vget_lane_f64, vrndn_f32,
+    vrndn_f64,
+};
+
+pub fn rint(x: f64) -> f64 {
+    // SAFETY: only requires target_feature=neon, ensured by `cfg_if` in parent module.
+    let x_vec: float64x1_t = unsafe { vdup_n_f64(x) };
+
+    // SAFETY: only requires target_feature=neon, ensured by `cfg_if` in parent module.
+    let result_vec: float64x1_t = unsafe { vrndn_f64(x_vec) };
+
+    // SAFETY: only requires target_feature=neon, ensured by `cfg_if` in parent module.
+    let result: f64 = unsafe { vget_lane_f64::<0>(result_vec) };
+
+    result
+}
+
+pub fn rintf(x: f32) -> f32 {
+    // There's a scalar form of this instruction (FRINTN) but core::arch doesn't expose it, so we
+    // have to use the vector form and drop the other lanes afterwards.
+
+    // SAFETY: only requires target_feature=neon, ensured by `cfg_if` in parent module.
+    let x_vec: float32x2_t = unsafe { vdup_n_f32(x) };
+
+    // SAFETY: only requires target_feature=neon, ensured by `cfg_if` in parent module.
+    let result_vec: float32x2_t = unsafe { vrndn_f32(x_vec) };
+
+    // SAFETY: only requires target_feature=neon, ensured by `cfg_if` in parent module.
+    let result: f32 = unsafe { vget_lane_f32::<0>(result_vec) };
+
+    result
+}
--- a/etc/function-definitions.json
+++ b/etc/function-definitions.json
@@ -604,12 +604,14 @@
     "rint": {
         "sources": [
             "src/libm_helper.rs",
+            "src/math/arch/wasm32.rs",
             "src/math/rint.rs"
         ],
         "type": "f64"
     },
     "rintf": {
         "sources": [
+            "src/math/arch/wasm32.rs",
             "src/math/rintf.rs"
         ],
         "type": "f32"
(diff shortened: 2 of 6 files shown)


src/math/arch/aarch64.rs
hanna-kruppe, line 1: For the record: there's a bunch of other libm functions that could be added here (sqrt, fma, probably floor/ceil/trunc as well) but I won't have time to work on that.

tgross35:
Thanks for jumping on this, the wasm changes look good to me. For aarch64, is there any reason not to just use assembly to get the non-vector version? For reference https://github.com/kraj/musl/blob/eb4309b142bb7a8bdc839ef1faf18811b9ff31c8/src/math/aarch64/rint.c

tgross35:
Also just as a note, wasm checks the build but doesn't actually run any tests. I am not sure of a good way to do so, but am open to ideas if you have any.

src/math/arch/aarch64.rs
tgross35: Eventually I will probably attempt to script together the operations from https://github.com/kraj/musl/tree/eb4309b142bb7a8bdc839ef1faf18811b9ff31c8/src/math/aarch64 (and the other architectures), most of the rounding-related ops are covered there. But yeah, it isn't a priority for the time being.

hanna-kruppe:
> Thanks for jumping on this, the wasm changes look good to me. For aarch64, is there any reason not to just use assembly to get the non-vector version? For reference https://github.com/kraj/musl/blob/eb4309b142bb7a8bdc839ef1faf18811b9ff31c8/src/math/aarch64/rint.c

I don't think there's any fundamental obstacle (I though MSRV might be a problem but apparently not), I just don't fully trust myself to write that inline asm correctly and the only cost from using the vector variant is an extra vdup, which is very cheap:

```
$ cargo asm --target aarch64-unknown-linux-gnu --features unstable-intrinsics -p libm --lib rintf
.section .text.libm::math::rintf::rintf,"ax",@progbits
        .globl  libm::math::rintf::rintf
        .p2align        2
        .type   libm::math::rintf::rintf,@function
libm::math::rintf::rintf:
        .cfi_startproc
        dup v0.2s, v0.s[0]
        frintn v0.2s, v0.2s
        ret
```



> Also just as a note, wasm checks the build but doesn't actually run any tests. I am not sure of a good way to do so, but am open to ideas if you have any.

You can run `cargo test` by installing wasmtime and setting `target.<triple>.runner = "wasmtime"` in the Cargo configuration (`.cargo/config.toml` or environment variable). I did this manually and it passed except for `check_coverage::ensure_list_is_updated` because it can't shell out. I don't know how this interacts with the other tests not run via `cargo test`.

tgross35 (approved):

tgross35:
> I don't think there's any fundamental obstacle (I though MSRV might be a problem but apparently not), I just don't fully trust myself to write that inline asm correctly and the only cost from using the vector variant is an extra vdup, which is very cheap. (I don't think the scalar vs. vector form makes any difference for the performance of rounding operation itself, at least as long as you stick to 64 bit vectors.)

For anything where MSRV would be a problem, I have just been gating things under `unstable_intrinsics` so at least anybody using this via std will benefit. There are some things that could be more fine grained, but I don't think it's worth adding more to possible feature combinations.

> You can run `cargo test` by installing wasmtime and setting `target.<triple>.runner = "wasmtime"` in the Cargo configuration (`.cargo/config.toml` or environment variable). I did this manually and it passed except for `check_coverage::ensure_list_is_updated` because it can't shell out. I don't know how this interacts with the other tests not run via `cargo test`.

I do need to document this better but the only tests enabled by default are a handful of spotchecks, it needs the `build-musl` feature to build musl and compare against that, and/or `test-multiprecision` to test against MPFR. I have my doubts that either of those are likely work on wasm, unfortunately.

Thanks for the additions here!

hanna-kruppe:
For what it's worth, wasi-libc uses a lightly modified musl for its libm and is used by Rust's wasm32-wasip1 target as part of `std`, so it might be relatively easy to compare against that. Though it might not be very useful for this specific change because they probably just end up using the same lowering for any functions that map directly to wasm instructions.

PR #442, opened 2025-01-14
Simplify and optimize fdim

quaternic:
The cases with `NaN` arguments can be handled by the same `x - y` expression, and this generates much better code: 
https://godbolt.org/z/r6sa6o5ov

### DIFF
--- a/src/math/generic/fdim.rs
+++ b/src/math/generic/fdim.rs
@@ -1,13 +1,9 @@
 use super::super::Float;
 
 pub fn fdim<F: Float>(x: F, y: F) -> F {
-    if x.is_nan() {
-        x
-    } else if y.is_nan() {
-        y
-    } else if x > y {
-        x - y
-    } else {
+    if x <= y {
         F::ZERO
+    } else {
+        x - y
     }
 }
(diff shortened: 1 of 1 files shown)


tgross35 (approved):
Great find! Looks like it needs a format run.

src/math/generic/fdim.rs
```diff
@@ -1,13 +1,9 @@
...
-    } else if y.is_nan() {
-        y
-    } else if x > y {
-        x - y
-    } else {
+    if x <= y {
         F::ZERO
+    } else {
+        x - y
     }
```
quaternic, line 8:
```suggestion
    if x <= y { F::ZERO } else { x - y }
```

tgross35:
This eliminates a couple of libcalls for `f16` and `f128` as well https://godbolt.org/z/cf6e3jY9G

quaternic:
From your godbolt I noticed I had posted a link with incorrect code (`x < y` instead of `x <= y`). Updated the OP with the correct version of each comparison. (But the comparison looks essentially the same.)

PR #443, opened 2025-01-14
[experiment] only test `libm-test` with the release profile

tgross35:


tgross35:
No speedups, and I forgot we need to run tests in `libm` anyway.

PR #453, opened 2025-01-22
Shift then mask, rather than mask then shift

tgross35:
This may allow for small optimizations with larger float types since `u32` math can be used after shifting. LLVM may be already getting this anyway.


tgross35:
I am going to wait until the floor and ceil PRs merge so the difference, if any, stands out in benchmarks.

PR #457, opened 2025-01-22
Add `hf16!` and `hf128!`

tgross35:
Expand the existing hex float functions and macros with versions that work with `f16` and `f128`.

Cc @quaternic since this is your code

### DIFF
--- a/src/math/support/hex_float.rs
+++ b/src/math/support/hex_float.rs
@@ -4,6 +4,12 @@
 
 use super::{f32_from_bits, f64_from_bits};
 
+/// Construct a 16-bit float from hex float representation (C-style)
+#[cfg(f16_enabled)]
+pub const fn hf16(s: &str) -> f16 {
+    f16::from_bits(parse_any(s, 16, 10) as u16)
+}
+
 /// Construct a 32-bit float from hex float representation (C-style)
 pub const fn hf32(s: &str) -> f32 {
     f32_from_bits(parse_any(s, 32, 23) as u32)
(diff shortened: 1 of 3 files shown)


src/math/support/hex_float.rs
```diff
@@ -308,16 +365,62 @@ mod tests {
...
-        assert_eq!(hf32!("0x1.ffep+8").to_bits(), 0x43fff000u32);
-        assert_eq!(hf64!("0x1.ffep+8").to_bits(), 0x407ffe0000000000u64);
+        // FIXME(msrv): enable once parsing works
+        // #[cfg(f16_enabled)]
+        // assert_eq!(hf16!("0x1.ffp+8").to_bits(), 0x5ffc_u16);
+        assert_eq!(hf32!("0x1.ffep+8").to_bits(), 0x43fff000_u32);
+        assert_eq!(hf64!("0x1.ffep+8").to_bits(), 0x407ffe0000000000_u64);
+        // FIXME(msrv): enable once parsing works
+        // #[cfg(f128_enabled)]
+        // assert_eq!(hf128!("0x1.ffep+8").to_bits(), 0x4007ffe0000000000000000000000000_u128);
```
quaternic, line 423: Should these be uncommented?

src/math/support/hex_float.rs
```diff
@@ -14,6 +19,12 @@ pub const fn hf64(s: &str) -> f64 {
     f64_from_bits(parse_any(s, 64, 52) as u64)
 }
 
+/// Construct a 64-bit float from hex float representation (C-style)
```
quaternic, line 22:
```suggestion
/// Construct a 128-bit float from hex float representation (C-style)
```

src/math/support/hex_float.rs
```diff
@@ -4,6 +4,11 @@
 
 use super::{f32_from_bits, f64_from_bits};
 
+#[cfg(f16_enabled)]
```
quaternic, line 7:
```suggestion

/// Construct a 16-bit float from hex float representation (C-style)
#[cfg(f16_enabled)]
```

quaternic (commented):
A couple of comments need fixing, but overall looks good to me!

I'm not sure if I was considering f128 when writing the parsing, so I read through it again just to check that there shouldn't be anything where it might unexpectedly overflow.

src/math/support/hex_float.rs
tgross35: They couldn't be uncommented here because the MSRV test would complain even though it wasn't enabled (looks like this was changed at 1.67 https://rust.godbolt.org/z/7dWP6jTTE). But the macro trick works, so I just moved it there.

tgross35:
Thanks for taking a look!

PR #461, opened 2025-01-23
Test with the unreleased version of rug

tgross35:
Adds `frexp` and `remquo` that we can use directly.


tgross35:
The new version was released, will be upgraded with https://github.com/rust-lang/libm/pull/477.

PR #464, opened 2025-01-23
Add a generic version of `scalbn`

tgross35:
This replaces the `f32` and `f64` versions of `scalbn` and `ldexp`.


tgross35:
This does not yet work for `f16` or `f128`, so splitting off from https://github.com/rust-lang/libm/pull/391.

tgross35:
No change for `scalbn`, slight 3% icount regression for `scalbnf`. This is tolerable.

PR #469, opened 2025-01-24
Add `fmodf16` and `fmodf128`

tgross35:


tgross35:
`fmodf128` seems to be unbearably slow (as in, the slowest thing we test) so I am splitting that off until I can look into it further https://github.com/rust-lang/libm/pull/470.

src/math/generic/fmod.rs
```diff
@@ -0,0 +1,84 @@
...
+    if ix << 1 <= iy << 1 {
+        if ix << 1 == iy << 1 {
+            return F::ZERO * x;
+        }
+        return x;
+    }
+
+    /* normalize x and y */
+    if ex == 0 {
+        let i = ix << F::EXP_BITS;
```
sjrd, line 29: I think this was mistranslated from the previous loop. It fails to account for the sign bit. I believe it should be
```suggestion
        let i = ix << (F::EXP_BITS + 1);
```
I discovered this because I translated this implementation to Wasm by hand for use in the Scala.js-to-Wasm backend:
https://github.com/scala-js/scala-js/pull/5147
My test cases with subnormals failed with the formula as is, but do pass with the added `+ 1`.

src/math/generic/fmod.rs
```diff
@@ -0,0 +1,84 @@
...
+        ex -= i.leading_zeros() as i32;
+        ix <<= -ex + 1;
+    } else {
+        ix &= F::Int::MAX >> F::EXP_BITS;
```
sjrd, line 33: There is something similar here. I believe the *intent* was also to have the `+ 1` :
```suggestion
        ix &= F::Int::MAX >> (F::EXP_BITS + 1);
```
However in this branch, it doesn't make any difference. The only bit for which it can change the result is anyway or'ed back to 1 on the very next line. But it seems more clearly correct with the `+ 1`, as it then creates a mask that exactly matches the mantissa bits, rather than "the mantissa bits plus arbitrarily the low-order bit of the exponent".

src/math/generic/fmod.rs
tgross35: Thanks for finding this, I think you are correct. I'll make sure the failures get tested with the fix. I'll double check with all the cases from https://github.com/scala-js/scala-js/pull/5147, but do you happen to know off the top of your head which ones failed?

sjrd: Among the tests I wrote, the following fail without the `+ 1`s.

Tests are `test(expectedResult, x, y)`:

For `Double` (binary64):
```
test(2.0696e-320, 2.1, 3.123e-320)
test(1.772535e-318, 2.1, 2.253547e-318)
```
For `Float` (binary32):
```
test(8.085e-42f, 2.1f, 8.858e-42f)
test(6.1636e-40f, 2.1f, 6.39164e-40f)
test(4.77036e-40f, 5.5f, 6.39164e-40f)
test(-5.64734e-40f, -151.189f, 6.39164e-40f)
```
Basically they're cases where `x` is a normal but `y` is a subnormal. If they're both subnormals, the two errors "cancel out". And if `x` is a subnormal but `y` is a normal, then `x < y` and so the earlier branch exits early before getting here.

tgross35: Thanks for the detailed info, opened https://github.com/rust-lang/libm/pull/535 to look at this.

PR #473, opened 2025-01-24
Split a redundant condition in `fmod`

tgross35:
`i >> n == 0` implies `i == 0` so we don't need to check both. Breaking this out eliminates two labels and removes a few instructions in `fmod` on my machine.


tgross35:
Interesting, I checked the asm on an aarch64 machine so maybe this makes things worse on x86 (or maybe the visual improvements to assembly didn't actually reduce anything).

Will rerun once we have a f128 baseline.

PR #476, opened 2025-01-25
Remove `#[allow(...)]` attributes that are no longer needed

tgross35:


tgross35:
Looks like there is still a lot that we don't yet use.

PR #479, opened 2025-01-25
Rework the available Cargo profiles

tgross35:
Currently the default release profile enables LTO and single CGU builds, which is very slow. Most tests are better run with optimizations enabled since it allows testing a much larger number of inputs, so it is inconvenient that building can sometimes take significantly longer than the tests.

Remedy this by doing the following:

* Move the existing `release` profile to `release-opt`.
* With the above, the default `release` profile is untouched (16 CGUs and thin local LTO).
* `release-checked` inherits `release`, so no LTO or single CGU.

This means that the simple `cargo test --release` becomes much faster for local development. We are able to enable the other profiles as needed in CI.

Tests should ideally still be run with `--profile release-checked` to ensure there are no debug assetions or unexpected wrapping math hit.

`no-panic` still needs a single CGU, so must be run with `--profile release-opt`. Since it is not possible to detect CGU or profilel configuration from within build scripts, the `ENSURE_NO_PANIC` environment variable must now always be set.

ci: allow-regressions

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -61,18 +61,20 @@ exclude = [
 [dev-dependencies]
 no-panic = "0.1.33"
 
-[profile.release]
-# Options for no-panic to correctly detect the lack of panics
-codegen-units = 1
-lto = "fat"
+# The default release profile is unchanged.
 
 # Release mode with debug assertions
 [profile.release-checked]
-codegen-units = 1
-debug-assertions = true
 inherits = "release"
-lto = "fat"
+debug-assertions = true
 overflow-checks = true
+
+# Release with maximum optimizations, which is very slow to build. This is also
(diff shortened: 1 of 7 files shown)


tgross35:
CI time is consistent, the wins from not LTOing every test probably washes out with whatever performance gain there could have been.

tgross35:
Intereesting that the sanity check failed here, I will update with your suggestion when I get the chance. The random tests occasionally fail for atan2 and the bessel functions, but I never checked to see if it was reproducible.

tgross35:
CI on some other platforms is now stopping at https://github.com/rust-lang/rust/issues/136096. I will just wait for https://github.com/rust-lang/rust/pull/136098 to land.

PR #482, opened 2025-01-28
Test whether core-math can be built on more platforms

tgross35:
We can't build easily build musl or MPFR on Windows or a few other platforms. Maybe core-math is easier to build.


tgross35:
Looks like `__int128` support plus some `printf` differences on Windows mean this might not be any easier.

PR #483, opened 2025-01-28
Add support for printing hex float syntax

tgross35:
Update the `hf*` functions to be able to parse infinities and NANs, then add a wrapper type that formats with the hex float syntax.

For now this is only being used in `util`, but in a followup I will somehow merge this into the existing `Hex` trait so this is displayed from test output.

### DIFF
--- a/src/math/support/hex_float.rs
+++ b/src/math/support/hex_float.rs
@@ -2,7 +2,9 @@
 
 #![allow(dead_code)] // FIXME: remove once this gets used
 
-use super::{f32_from_bits, f64_from_bits};
+use core::fmt;
+
+use super::{Float, f32_from_bits, f64_from_bits};
 
 /// Construct a 16-bit float from hex float representation (C-style)
 #[cfg(f16_enabled)]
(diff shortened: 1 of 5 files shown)


tgross35:
Cc @quaternic, more parsing and printing

tgross35:
Looks like indicatif needs a temporary pin https://github.com/rust-lang/libm/pull/484

src/math/support/hex_float.rs
```diff
@@ -2,41 +2,47 @@
...
 #[cfg(f128_enabled)]
 pub const fn hf128(s: &str) -> f128 {
-    f128::from_bits(parse_any(s, 128, 112))
+    f128::from_bits(parse_any(s, 128, 112, f128::NAN.to_bits()))
 }
 
-const fn parse_any(s: &str, bits: u32, sig_bits: u32) -> u128 {
+/// Parse any float from hex to its bitwise representation.
+///
+/// `nan_repr` is passed rather than constructed so the platform-specific NaN is returned.
```
quaternic, line 31: Is it necessary to return specifically `fN::NAN`? Rather than, say, the quiet NaN with 0 payload.

tgross35: I had this for bitwise equality at some point, but I since removed that path. I'll clean this up.

crates/libm-test/src/num.rs
```diff
@@ -267,6 +268,106 @@ pub fn linear_ints(
...
+    // The width is rounded up to the nearest char (4 bits)
+    let mwidth = (F::SIG_BITS as usize + 3) / 4;
+    let leading = if exponent == -bias {
+        // subnormal number means we shift our output by 1 bit.
+        exponent += 1;
+        "0."
+    } else {
+        "1."
+    };
+    let exp_sign = if exponent >= 0 { "+" } else { "" };
```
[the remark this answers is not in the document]
tgross35: 🤯 I don't think I ever knew about that option, thanks

src/math/support/hex_float.rs
tgross35: Removed. I also moved the new struct to `hex_float` so it is usable for debug printing within `libm`.

PR #493, opened 2025-02-06
fma: Ensure zero has the correct sign

tgross35:
Currently, `fma(tiny, -tiny, 0.0)` returns 0.0 while the answer should be -0.0. This is because `-0.0 + 0.0 = +0.0` in the default rounding mode; however, the result should be negative. Musl has the same pattern but that version worked because the C compiler was contracting `x*y + z` to (ironically) `fmadd`.

Musl was fixed in 9683bd6241 ("math: fix fma(x,y,0) when x*y rounds to -0"). Add the same fix here, which allows dropping the xfails.

Also includes a submodule update of musl that has the fix.


tgross35:
Switching to the `.is_finite()` version was unfortunately less performant, so sticking with the comparison to `ZEROINFNAN` for now.

PR #503, opened 2025-02-08
Implement `u256` with two `u128`s rather than `u64`

tgross35:
This produces better assembly, e.g. on aarch64:

            .globl  libm::u128_wmul
            .p2align        2
    libm::u128_wmul:
    Lfunc_begin124:
            .cfi_startproc
            mul x9, x2, x0
            umulh x10, x2, x0
            umulh x11, x3, x0
            mul x12, x3, x0
            umulh x13, x2, x1
            mul x14, x2, x1
            umulh x15, x3, x1
            mul x16, x3, x1
            adds x10, x10, x14
            cinc x13, x13, hs
            adds x13, x13, x16
            cinc x14, x15, hs
            adds x10, x10, x12
            cinc x11, x11, hs
            adds x11, x13, x11
            stp x9, x10, [x8]
            cinc x9, x14, hs
            stp x11, x9, [x8, #16]
            ret

The original was ~70 instructions so the improvement is significant. With these changes, the result is reasonably close to what LLVM generates using `u256` operands [1].

[1]: https://llvm.godbolt.org/z/re1aGdaqY

PR #505, opened 2025-02-08
Change how operators are `black_box`ed

tgross35:
For some reason, the upcoming limb changes in [1] seem to ignore the black boxing when applied to the operator function. Changing to instead black box the inputs appears to fix this.

I am merging this in advance so our baseline is accurate.

[1]: https://github.com/rust-lang/libm/pull/503

ci: allow-regressions


tgross35:
Thanks, the reasoning makes sense. Most of our other cases should be fine then since they go through the function pointer `black_box(Op::ROUTINE)`, correct?

That behavior difference is subtle, it seems worth a rustc issue in case there is something better to be done there.

PR #523, opened 2025-02-24
Make the compiler-builtins test more accurately mirror compiler-builtins

tgross35:
In `compiler-builtins`, `libm` is contained within a `math` module. The smoke test in this repo has a slightly different layout so some things were passing that shouldn't be.

Change module layouts in `compiler-builtins-smoke-test` to match `compiler-builtins` and update a few instances of broken paths.


tgross35:
Added a CI patch like https://github.com/rust-lang/compiler-builtins/commit/b2bcfc838e2a4b72fa62b333e3eb91f250aa4539 since the same issues are popping up here.

PR #529, opened 2025-03-03
Rounding mode support for hexadecimal float parsing, with improvements to error handling

quaternic:
Parsing errors are now bubbled up part of the way, but that needs some more work.

Rounding should be correct, and the `Status` returned by `parse_any` should have the correct bits set. These are used for the current (unchanged) behaviour of the surface level functions like `hf64`: panic on invalid inputs, or values that aren't exactly representable.

### DIFF
--- a/src/math/support/hex_float.rs
+++ b/src/math/support/hex_float.rs
@@ -2,149 +2,240 @@
 
 use core::fmt;
 
-use super::{Float, f32_from_bits, f64_from_bits};
+use super::{f32_from_bits, f64_from_bits, Float, Round, Status};   
 
 /// Construct a 16-bit float from hex float representation (C-style)
 #[cfg(f16_enabled)]
 pub const fn hf16(s: &str) -> f16 {
-    f16::from_bits(parse_any(s, 16, 10) as u16)
+    match parse_any(s, 16, 10, Round::Nearest) {
+        (bits, Status::OK) => f16::from_bits(bits as u16),
+        (_, status) => status_panic(status),
+    }
 }
 
 /// Construct a 32-bit float from hex float representation (C-style)
 #[allow(unused)]
 pub const fn hf32(s: &str) -> f32 {
-    f32_from_bits(parse_any(s, 32, 23) as u32)
+    match parse_any(s, 32, 23, Round::Nearest) {
+        (bits, Status::OK) => f32_from_bits(bits as u32),
+        (_, status) => status_panic(status),
+    }
 }
 
 /// Construct a 64-bit float from hex float representation (C-style)
 pub const fn hf64(s: &str) -> f64 {
-    f64_from_bits(parse_any(s, 64, 52) as u64)
+    match parse_any(s, 64, 52, Round::Nearest) {
+        (bits, Status::OK) => f64_from_bits(bits as u64),
+        (_, status) => status_panic(status),
+    }
 }
 
 /// Construct a 128-bit float from hex float representation (C-style)
 #[cfg(f128_enabled)]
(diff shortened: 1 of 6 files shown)


src/math/support/hex_float.rs
```diff
@@ -2,149 +2,258 @@
...
-
-    // the most significant bit is like having 1 in the exponent bits
-    // add any leftover exponent to that
-    assert!(uexp >= 0 && uexp < (1 << exp_bits) - 2);
-    sig += (uexp as u128) << sig_bits;
+}
 
-    // finally, set the sign bit if necessary
-    sig | ((neg as u128) << (bits - 1))
+/// Divide by 4, rounding accor
```
tgross35, line 201: Did the rest of this get cut off?

tgross35 (approved):
Sorry I disappeared for a bit soon after you posted this, but it looks great! Just one possible doc fix.

src/math/support/hex_float.rs
quaternic: Yeah. The cutting off happened when I couldn't think of a good wording while writing it and moved to something else and didn't notice it later.
```suggestion
/// Divide by 4, rounding with the given mode
```

quaternic:
@tgross35 
> Sorry I disappeared for a bit soon after you posted this, but it looks great! Just one possible doc fix.
Thanks! I'm sorry for taking my time as well. I fixed the comment and rebased; and CI seems to still be happy with it. Could you squash this while merging?

CrazyboyQCD:
There are some magic numbers and literals, it would be good to make them constants for better readability.

src/math/support/hex_float.rs
```diff
@@ -2,149 +2,258 @@
...
+}
 
-    // finally, set the sign bit if necessary
-    sig | ((neg as u128) << (bits - 1))
+/// Divide by 4, rounding with the given mode
+const fn shr2_round(mut x: u128, round: Round) -> u128 {
+    let t = (x as u32) & 0b111;
+    x >>= 2;
+    match round {
+        Round::Nearest => x + ((0b11001000_u8 >> t) & 1) as u128,
```
quaternic, line 206: @CrazyboyQCD If you meant this, yeah, it could use an explanatory comment. I don't think extracting to a constant would help since it only makes sense in the context it is used in. ( Though this and the previous function could be more generally useful as utilities for `Round`. )
```suggestion
        // Look-up-table on the last three bits for when to round up 
        Round::Nearest => x + ((0b11001000_u8 >> t) & 1) as u128,
        
```
Or was there something else?

tgross35: I just applied this since it's the only case I see, @CrazyboyQCD feel free to submit a followup PR or ask for clarification if there is anything else.

src/math/support/hex_float.rs
```diff
@@ -2,149 +2,260 @@
...
-    sig | ((neg as u128) << (bits - 1))
+/// Divide by 4, rounding with the given mode
+const fn shr2_round(mut x: u128, round: Round) -> u128 {
+    let t = (x as u32) & 0b111;
+    x >>= 2;
+    match round {
+        // Look-up-table on the last three bits for when to round up 
+        Round::Nearest => x + ((0b11001000_u8 >> t) & 1) as u128,
+        
+        Round::Negative => x,
```
tgross35, line 209: Rustfmt

```suggestion
        // Look-up-table on the last three bits for when to round up
        Round::Nearest => x + ((0b11001000_u8 >> t) & 1) as u128,

        Round::Negative => x,
```

PR #532, opened 2025-04-05
Reenable the use of the force-soft-floats feature

antoyo:
This fixes an infinite recursion in sqrt for codegen backends that implement intrinsics like simd_fsqrt by calling sqrt on every element of the vector.

This is going to be needed to fix https://github.com/rust-lang/compiler-builtins/issues/649.

I was able to test this locally by changing some other code (basically having an implementation of `_mm_sqrt_pd` that calls the local `sqrt` function on all elements of the vector).
Do you have any ideas for how to add a test for this? Perhaps by including this implementation of `_mm_sqrt_pd` behind a `#[cfg]`?

### DIFF
--- a/crates/libm-test/src/precision.rs
+++ b/crates/libm-test/src/precision.rs
@@ -106,11 +106,10 @@ pub fn default_ulp(ctx: &CheckCtx) -> u32 {
     }
 
     if cfg!(target_arch = "x86") {
-        match ctx.fn_ident {
-            // Input `fma(0.999999999999999, 1.0000000000000013, 0.0) = 1.0000000000000002` is
-            // incorrect on i586 and i686.
-            Id::Fma => ulp = 1,
-            _ => (),
+        // Input `fma(0.999999999999999, 1.0000000000000013, 0.0) = 1.0000000000000002` is
+        // incorrect on i586 and i686.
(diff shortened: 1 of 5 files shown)


tgross35:
I think this PR shouldn't be needed; usage of `force-soft-floats` in source was replaced with `arch_enabled` to simplify the logic, which `select_implementation!` already guards against https://github.com/rust-lang/libm/blob/c9672e5a1a75bfa82981b6240b7bc3ed3524b8b3/src/math/support/macros.rs#L74-L75. The config is set here https://github.com/rust-lang/libm/blob/c9672e5a1a75bfa82981b6240b7bc3ed3524b8b3/configure.rs#L79-L88, `compiler-builtins` probably just needs to be fixed to have the same logic.

antoyo:
> I think this PR shouldn't be needed; usage of `force-soft-floats` in source was replaced with `arch_enabled` to simplify the logic, which `select_implementation!` already guards against

I thought `arch_enabled` would enable the usage of assembly. So won't this disable assembly in addition to hardware floats?

tgross35:
(replying at https://github.com/rust-lang/compiler-builtins/pull/810 since I had a better answer there)

antoyo:
Closing since this was fixed elsewhere.

PR #535, opened 2025-04-15
fmod: add regression tests for subnormal issue

tgross35:
From discussion at [1] our loop count calculation is incorrect, causing an issue with subnormal numbers. Add test cases for known failures.

[1]: https://github.com/rust-lang/libm/pull/469#discussion_r2012473920


tgross35:
@sjrd I added the test cases mentioned at https://github.com/rust-lang/libm/pull/469#discussion_r2022337272, thank you for providing those. It doesn't seem like I am able to reproduce the failure without the `+1`. Is it possible there is a subtle difference in the port? Spot checking, the results for this crate, musl, and mpfr all seem to line up.

```text
$ cargo run -p util -- eval musl fmodf 2.1 0x1.8b1p-137
8.085e-42 0x0.002d14p-126
$ cargo run -p util -- eval mpfr fmodf 2.1 0x1.8b1p-137
8.085e-42 0x0.002d14p-126
$ cargo run -p util -- eval libm fmodf 2.1 0x1.8b1p-137
8.085e-42 0x0.002d14p-126
```

tgross35:
Cc @quaternic who has done a lot of experimentation with fmod

quaternic:
Right, the translation to a generic implementation introduced an unintentional change that subnormals are "normalized" to a form that is `(2 * N) * 2^ (E - 1)` instead of the intended `(N * 2^E)`, where `N.ilog2() == SIG_BITS`. The overall value is still the same, so probably the code that follows just happens to work correctly for all those cases. That likely hinges on various subtle details and it's not surprising that the Scala-port wouldn't have done *everything* exactly the same.

sjrd:
One thing we did differently in the Scala port is that we mask off the sign bit early:
https://github.com/scala-js/scala-js/pull/5149/files#diff-cb80a0a9c84409f42505687656241e73f7aa71519db22f7cfc6be9f6b0751c75R83-R88
We do this because we don't have (native) unsigned comparisons. With the sign bit masked off, we can use signed comparisons with the same outcome.

Other than that, I followed the Rust implementation pretty closely. Perhaps the `saturating_sub` towards the end makes an actual difference? (I used `wrapping_sub` here as well, since that's all we have natively in Scala, again.)

tgross35:
It's possible there is a difference if signed arithmetic is used rather than unsigned at some point since the net effect is to shift one bit further left than expected, which would be a sign flip any of the subtract+shift ops wind up placing that in the MSB. I don't think it's worth digging into too much though; I'll apply the change here since it's more correct, but this will go away soon with @quaternic's rewrite anyway https://github.com/rust-lang/libm/pull/536.

